PR #16, opened 2015-11-05
Allow strongly typed extendable filtering

niemyjski:
Previously all of our internal repository methods took an object and we casted them to an interface and preformed some action.. This works great if you have the source to the repo but not so great when using as a third party library..

This changes makes a query builder interface that allows you to build up a query and have strongly typed filtering.


src/Core/Repositories/Queries/IQueryBuilder.cs
```diff
@@ -0,0 +1,7 @@
+﻿using System;
+
+namespace Foundatio.Repositories.Queries {
+    public interface IQueryBuilder<T> where T : class, new() {
+        T Build(bool supportSoftDeletes = false);
+    }
```
niemyjski: This could be moved into the elastic repo. I just thought it was pretty generic.

src/Elasticsearch/Repositories/Query/ElasticQuery.cs
```diff
@@ -1,14 +0,0 @@
...
-
-namespace Foundatio.Elasticsearch.Repositories.Queries {
-    public class ElasticQuery : Query, IElasticFilterQuery, IElasticIndicesQuery {
-        public ElasticQuery() {
-            Indices = new List<String>();
-        }
-
-        public FilterContainer ElasticFilter { get; set; }
-        public List<string> Indices { get; set; }
-    }
```
niemyjski: These fields were merged into the base class elastic builder class.

src/Elasticsearch/Repositories/Query/Parts/SoftDeletesQuery.cs
```diff
@@ -1,14 +0,0 @@
-﻿using System;
-
```
niemyjski: This was a duplicate class.

niemyjski:
We could rename TQueryBuilder to TQuery..

PR #22, opened 2016-01-25
Added support for Azure Storage Queues

jamie94bc:
#21

Required alteration of queue API to support poison queue. Without poison queue this could be achieved by concatenating the message ID and pop receipt to form a composite ID.


src/Core/Queues/IQueue.cs
```diff
@@ -21,10 +21,16 @@ public interface IQueue<T> : IHaveSerializer, IDisposable where T : class {
 
         Task<QueueEntry<T>> DequeueAsync(CancellationToken cancellationToken = default(CancellationToken));
 
+        [Obsolete("Use QueueEntry<T> overload")]
```
niemyjski: Thanks for your pull request!!  Is there any chance you can revert these changes to the interface. I'm not remembering specifics, but we tried to go this route and ended up reverting to just take a string id.

ejsmith: Yes, thank you very much! I am going to try and remember why we changed the interfaces to just send the string ids. If I remember right it was just because we have seen other queues that wouldn't be able to implement the interface with the full object. I think it might be ServiceBus queues. Also, we did some stuff in QueueBase that could be helpful to you:

https://github.com/exceptionless/Foundatio/blob/master/src/Core/Queues/QueueBase.cs#L93

jamie94bc: Aha - should have looked in the base. The only issue with this is to have the poison queue functionality we need the [value of the queue entry](https://github.com/exceptionless/Foundatio/pull/22/files#diff-65336afcf3620424bd815d6ab2436d62R105) to push onto the queue. That would essentially require putting the entire QueueEntry into the cache!

niemyjski:
I wasn't sure what you meant by poison queue until I looked (`deadletter`). Can we just get the item from the queue and remove it and then place it in deadletter? instead of just keeping the entire entry around?

jamie94bc:
Yeah that's what I'm doing at the moment - but without the `AbandonAsync(QueueEntry<T> entry)` overload (or storing `QueueEntry.Value` in cache) all we'd be pushing to the deadletter queue would be the ID :cry:

ejsmith:
I guess Azure Storage Queues don't automatically support moving things to the deadletter queue after X retries?

jamie94bc:
Nope! http://www.c-sharpcorner.com/UploadFile/fe6121/deep-dive-into-azure-storage-queue-vs-azure-service-bus-queu/

jamie94bc:
Here's Microsoft's take on it: https://azure.microsoft.com/en-gb/documentation/articles/websites-dotnet-webjobs-sdk-storage-queues-how-to/#poison

I'm not sure how much of a deal this is for the purpose of this PR as all the queue implementations bar `InMemoryQueue` have non implemented `GetDeadletterItemsAsync` methods (which in Azure Storage and Service Bus's case could only be implemented by reading and removing items from the entire deadletter or "poison" queue).

Also I don't see why using the `QueueEntry<T>` object for the Complete/Abandon methods would be problematic with the Service Bus implementation? If anything it would make more sense since Service Bus requires a `lockToken` to complete/abandon rather than the message ID. So currently if I called `ServiceBusQueue.EnqueueAsync` followed by `CompleteAsync` directly, that would fail since `EnqueueAsync` returns `message.MessageId` whereas `CompleteAsync` requires `message.LockToken`. Continuing on that theme, it might make sense to change the return type of `EnqueueAsync` from an id to the entry too?

ejsmith:
Sorry for the delay on this. I am going to check into this today.

ejsmith:
I am in the process of going through and doing a refactor to accommodate this change. I am mostly done, but I have plans today. Will be back on later tonight to try and finish it up. Then we can evaluate.

jamie94bc:
Refactor looks good - I didn't know how happy you'd be with breaking changes :smiley:

ejsmith:
Yeah, it should just be a recompile for most people. I think it's a little too early in the project to keep a bunch of deprecated methods around.

Still have some failing tests that I am looking into.

jamie94bc:
If it's tests for Azure storage queue it might be cancellation token related - when I used the `linkedCancellationToken` in the dequeue method it caused some tests to fail.

niemyjski:
Thanks, @ejsmith were you able to look into this?

ejsmith:
Yeah, passing the cancellation tokens in seem to make the tests hang. Would be good to figure out what is going on with that.

niemyjski:
@ejsmith @jamie94bc Is there anything I can do to progress this pr?

ejsmith:
Yeah, if you want to pull my changes and try and figure out why the tests are failing. Then that would be a huge help.

niemyjski:
@ejsmith where are your changes at?

ejsmith:
They are in the azurestoragequeues branch in this repository.

jamie94bc:
I've found the cause but not the solution for the hanging tests.

If you pass the cancellation token to the `GeMessageAsync` method of the Azure SDK, it causes a weird issue where the `linkedCancellationToken` in the `StartWorking` method never appears to be cancelled despite `_queueDisposedCancellationTokenSource.Token` showing as cancelled. The `linkedCancellationToken` works fine in the `DequeueAsync` method.

jamie94bc:
This fixes it so it's definitely async/context related. Moving `linkedCancellationToken` inside the `Task.Run(async () => {})` doesn't solve the issue.

```C#
var linkedCancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_queueDisposedCancellationTokenSource.Token, cancellationToken).Token;
Func<bool> isCancellationRequested = () => linkedCancellationToken.IsCancellationRequested;

Task.Run(async () => {
    while (!isCancellationRequested()) {
        IQueueEntry<T> queueEntry = null;
        try {
            queueEntry = await DequeueAsync(cancellationToken).AnyContext();
        }
        catch (TaskCanceledException) { }

        if (isCancellationRequested() || queueEntry == null)
            continue;

        try { 
            await handler(queueEntry, cancellationToken);
            if (autoComplete)
                await queueEntry.CompleteAsync().AnyContext();
        }
        catch (Exception ex) {
            Logger.Error().Exception(ex).Message("Worker error: {0}", ex.Message).Write();
            Interlocked.Increment(ref _workerErrorCount);
        }
    }
#if DEBUG
    Logger.Trace().Message("Worker exiting: {0} Cancel Requested: {1}", _queueReference.Name, linkedCancellationToken.IsCancellationRequested).Write();
#endif
}, linkedCancellationToken);
```

jamie94bc:
Interestingly this is also an issue when using the synchronous versions of `GetMessageAsync` without the `Func<bool> isCancellationRequested` hack.

niemyjski:
sigh, think we should log an issue with them to take a look?

ejsmith:
We can log an issue and not use the cancellation token for now. Are all of the other tests passing?

niemyjski:
I can't tell, as I'm reinstalling my github client that won't launch.. That's fine (we can do that, just put a todo there).

niemyjski:
@jamie94bc can you log this azure storage issue on there issue tracker and reference this pull request

niemyjski:
This seems to fix those tests except one in memory one.. seems like we have some azure service bus tests failing too

jamie94bc:
The mentioned test fix - I've simply bumped the delay from 10ms to 50ms to account for the latency of Azure Service Bus (~35ms according to MS). I don't think there's a better way to do this at the moment

PR #28, opened 2016-02-22
Feature/logging

ejsmith:


src/Azure/Foundatio.AzureServiceBus.csproj
```diff
@@ -9,8 +9,9 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Foundatio.AzureServiceBus</RootNamespace>
     <AssemblyName>Foundatio.AzureServiceBus</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
```
niemyjski: Lets update it to 4.5.2 (https://visualstudiogallery.msdn.microsoft.com/47bded90-80d8-42af-bc35-4736fdd8cd13) as that's the only supported runtime by ms right now and then reinstall all the packages..

src/Core/Queues/InMemoryQueue.cs
```diff
@@ -113,42 +107,38 @@ public class InMemoryQueue<T> : QueueBase<T> where T : class {
...
-#endif
+                _logger.Trace().Message("Waited for dequeue: {0}", sw.Elapsed.ToString()).Write();
             }
 
             if (_queue.Count == 0)
                 return null;
-#if DEBUG
-            Logger.Trace().Message("Dequeue: Attempt").Write();
-#endif
+
```
niemyjski: Guess were saying if you inject a logger you want log messages eh..

src/Azure/Foundatio.AzureServiceBus.csproj
ejsmith: Thought about that, but we are trying to be as downward compat as possible which is why we were sill on 4.5.0.  The only reason I went to 4.5.1 is because it was required for using the logging nuget package.

PR #29, opened 2016-03-01
Fixes bug where scoped cache keys are returned

jamie94bc:
Also updated test coverage :smiley:

Not sure how high the performance implications are for recreating a dictionary?

### DIFF
--- a/src/Core/Caching/ScopedCacheClient.cs
+++ b/src/Core/Caching/ScopedCacheClient.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Foundatio.Extensions;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Core/Caching/ScopedCacheClient.cs
```diff
@@ -1,13 +1,18 @@
...
 
 namespace Foundatio.Caching {
     public class ScopedCacheClient : ICacheClient {
+        private readonly string _keyPrefix;
+
         public ScopedCacheClient(ICacheClient client, string scope) {
             UnscopedCache = client ?? new NullCacheClient();
             Scope = scope;
+
+            _keyPrefix = scope != null ? String.Concat(scope, ":") : String.Empty;
```
niemyjski: @ejsmith any reason why we can't just set the Scope property directly?

jamie94bc: I don't think the consumer needs to know about the `:`? Would be strange to pass a scope into the ctor and have the `Scope` property return something (slightly) different.

niemyjski: fair enough :). I think we need to check the key for `IsNullOrWhiteSpace`and triming the scope because we'd be prefixing an empty scope if passed ` ` or ``.

PR #30, opened 2016-03-07
fix expiring in redis increment

derekpitt:
In LM we are using the ThrottlingLockProvider with redis cache client.. 
Our keys were not expiring because the lua script was never hitting the expire command.

Also, redis can only handle whole seconds when using `expire`, so I modified the the lua scripts to round it up if what you sent in is a fraction of a second.

I added some tests..  but they take a few seconds to run since redis can only expire after a second. Maybe they are not worth it? Let me know


niemyjski:
@derekpitt Can you please sign the cla :) The pr looks good

Issue #36, opened 2016-04-04
Create async initialization method for AzureServiceBusQueue.


moswald:
A lot of the `Microsoft.ServiceBus.NamespaceManager` methods called in the `AzureServiceBusQueue` constructor have `async` counterparts. Since constructors can't be `async`, I propose deprecating the current constructor (shakes fist at semver) and adding a new private one that does very little. Then introduce a new factory method: `static async Task<AzureServiceBusQueue>` factory method that does all the same work the current constructor does synchronously.

ejsmith:
Yeah, probably bad to be doing a bunch of heavy calls in the constructor. I think a generalized lazy initialization pattern would be good.

Do you want to take a stab at this and create a pull request for it?

moswald:
I was planning on starting it tomorrow when I got back to the office. :) I just wanted to get the conversation rolling before then.

ejsmith:
Awesome! Thank you.

jamie94bc:
This relates to #25

moswald:
Regarding #25, I'll take a look at having an optional parameter that makes the queue lazy - it will create and connect only on the first enqueue/dequeue.

niemyjski:
Awesome :) I wonder if we always need to setup the connections/subscriptions immediately. How would you know there is an item to dequeue in a distributed environment.

jamie94bc:
@niemyjski I think that would be handled by the `StartWorking()` method? I assume @moswald is thinking about adding something like a `EnsureQueueAsync()` method that each public API calls first.

moswald:
@niemyjski I would expect that if there's nothing to dequeue, it would just time out.

niemyjski:
@moswald Yeah! Not sure what I was thinking before (long night of playing quantum break). I think having this would be great and it looks like it would be super simple to implement.

niemyjski:
This has been fixed in master and will be released via nuget shortly

Issue #25, opened 2016-02-12
Queue Management Improvements


jamie94bc:
Our current implementation of `IQueue` has a [DeleteQueueAsync](https://github.com/exceptionless/Foundatio/blob/master/src/Azure/Queues/AzureServiceBusQueue.cs#L68) method which actually recreates the queue.

This clearly isn't expected behaviour (meaning we can't delete the queue easily in the test tear town) and in addition, we are also performing remote calls in the constructor of the queue which isn't ideal.

We could introduce an additional interface to manage the queues, although we'd have to specifically create every queue and inject it into the container - and the bootstrapper class therefore must ideally be async compatible.

```c#
public interface IQueueManager<TQueue> where TQueue : IQueue<T> {
    Task<TQueue> CreateQueueAsync(string queueName /* other params */);
    Task<bool> QueueExistsAsync(string queueName);
    Task DeleteQueueAsync(string queueName);
}
``` 

A big down side to this would be it would also make it more difficult to specify additional constructor params for the queue (as used in AzureStorageQueue to control poll interval).

niemyjski:
Yeah, I think that queue creation should happen automatically by the queues upon first use. Delete definitely shouldn't recreate the queue but maybe set a flag saying it needs to be recreated. We could just set some internals to null on delete and every method has an Ensure() method to ensure it's recreated. Thoughts?

ejsmith:
I actually had an interface together that was super similar to this before, but bailed out due to time constraints. I think I had just an object parameter for options on the CreateQueue method. That way you can pass any type of object in for configuration that the various implementations could cast and use how they wanted. I am going to try and dig that up and see if there were any good ideas.

moswald:
While I'm getting my hands dirty with #36, I can take a look at having the delete be more permanent.

niemyjski:
This has been fixed in master and will be released via nuget shortly

PR #37, opened 2016-04-07
Async ServiceBus Queue construction and permanent Queue deletes

moswald:
Addresses both #25 and #36, and as a bonus cleans up the tests a bit so that they can run in parallel.

### DIFF
--- a/src/Azure/Queues/AzureServiceBusQueue.cs
+++ b/src/Azure/Queues/AzureServiceBusQueue.cs
@@ -283,7 +283,20 @@ namespace Foundatio.Queues {
                         MaxDeliveryCount = retries + 1,
                         LockDuration = workItemTimeout
                     };
-                    await namespaceManager.CreateQueueAsync(queueDescription).AnyContext();
+
+                    try {
+                        await namespaceManager.CreateQueueAsync(queueDescription).AnyContext();
+                    }
+                    catch (MessagingEntityAlreadyExistsException) {
+                        // do nothing - something created the queue between the check for exist and the call to create
+                    }
+                    catch (MessagingException) {
+                        // this one is a little more sinister - something conflicted with the create, but we don't know what
+                        // in that case, we'll re-check to see if something else created it
+                        if (!await namespaceManager.QueueExistsAsync(queueName).AnyContext()) {
+                            throw;
+                        }
+                    }
                 } else {
                     queueDescription = await namespaceManager.GetQueueAsync(queueName).AnyContext();
 
--- a/src/Azure/Queues/AzureServiceBusQueue.cs
+++ b/src/Azure/Queues/AzureServiceBusQueue.cs
@@ -1,79 +1,149 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
-using Foundatio.Extensions;
+﻿using Foundatio.Extensions;
 using Foundatio.Logging;
 using Foundatio.Serializer;
 using Microsoft.ServiceBus;
 using Microsoft.ServiceBus.Messaging;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Foundatio.Queues {
     public class AzureServiceBusQueue<T> : QueueBase<T> where T : class {
         private readonly string _queueName;
         private readonly NamespaceManager _namespaceManager;
-        private readonly QueueClient _queueClient;
-        private QueueDescription _queueDescription;
+        private readonly Func<Task<QueueDescription>> _createQueue;
+        private readonly Func<string, QueueClient> _clientFactory;
+        private Action _disposeClient = () => { };
         private long _enqueuedCount;
         private long _dequeuedCount;
         private long _completedCount;
         private long _abandonedCount;
         private long _workerErrorCount;
-        private readonly int _retries;
-        private readonly TimeSpan _workItemTimeout = TimeSpan.FromMinutes(5);
 
+        [Obsolete("This constructor runs synchronously. Prefer the async Factory.Build method.")]
         public AzureServiceBusQueue(string connectionString, string queueName = null, int retries = 2, TimeSpan? workItemTimeout = null, bool shouldRecreate = false, RetryPolicy retryPolicy = null, ISerializer serializer = null, IEnumerable<IQueueBehavior<T>> behaviors = null, ILoggerFactory loggerFactory = null) : base(serializer, behaviors, loggerFactory) {
             _queueName = queueName ?? typeof(T).Name;
             _namespaceManager = NamespaceManager.CreateFromConnectionString(connectionString);
-            _retries = retries;
-            if (workItemTimeout.HasValue && workItemTimeout.Value < TimeSpan.FromMinutes(5))
-                _workItemTimeout = workItemTimeout.Value;
+
+            workItemTimeout = workItemTimeout ?? TimeSpan.FromMinutes(5);
+            if (workItemTimeout.Value > TimeSpan.FromMinutes(5))
+                workItemTimeout = TimeSpan.FromMinutes(5);
+
+            _createQueue = MemoizeQueueFactory(_namespaceManager, queueName, workItemTimeout.Value, retries);
 
             if (_namespaceManager.QueueExists(_queueName) && shouldRecreate)
                 _namespaceManager.DeleteQueue(_queueName);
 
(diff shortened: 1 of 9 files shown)


src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -1,79 +1,149 @@
...
+                Timeout(workItemTimeout ?? TimeSpan.FromMinutes(5));
+                _shouldRecreate = shouldRecreate;
+                _lazyConnect = lazyConnect;
+                _retryPolicy = retryPolicy;
+                _serializer = serializer;
+                _behaviors = behaviors;
+                _loggerFactory = loggerFactory;
+            }
+
+            public Factory Queue(string queueName) {
```
jamie94bc: What's wrong with properties?

src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -1,79 +1,149 @@
...
+            private ISerializer _serializer;
+            private IEnumerable<IQueueBehavior<T>> _behaviors;
+            private ILoggerFactory _loggerFactory;
+
+            public Factory(string connectionString, string queueName = null, int retries = 2, TimeSpan? workItemTimeout = null, bool shouldRecreate = false, bool lazyConnect = true, RetryPolicy retryPolicy = null, ISerializer serializer = null, IEnumerable<IQueueBehavior<T>> behaviors = null, ILoggerFactory loggerFactory = null) {
+                _connectionString = connectionString;
+                Queue(queueName);
+                _retries = retries;
+                Timeout(workItemTimeout ?? TimeSpan.FromMinutes(5));
+                _shouldRecreate = shouldRecreate;
```
jamie94bc: I don't think we need to preserve this functionality (was implemented originally to be consistent with existing behaviour of other queues) - but if we're removing it, we should add a `ClearAsync()` method or similar to all queues.

src/Azure/Queues/AzureServiceBusQueue.cs
moswald: The methods allow a fluent interface:

````
var queue = new AzureServiceBusQueue.Factory(connectionString)
    .Queue(queueName)
    .Retries(0)
    .Build();
```

jamie94bc: Same as this, I don't think using a fluent interface has any benefits here but it's a personal preference I suppose.
```c#
var queue = new AzureServiceBusQueue.Factory(connectionString) {
   QueueName = queueName,
   Retries = 0
}.Build();
```

I wonder whether a factory is necessary - I'd be happy with something like:

```c#
AzureServiceBusQueue.CreateAsync(/* args */);
// or
AzureServiceBusQueue.CreateAsync(new AzureQueueCreationOptions { /* props */ })
```

moswald: Yeah, either works for me. I agree the properties version looks better. I'll change it.

jamie94bc:
Speaking to @niemyjski, I think what might be better is to declare both a `IQueue<T>` and `IQueueManager` interface that each of our implementations implement, the latter explicitly. The lazy creation should still be there.

For example...
```c#
public interface IQueueManager {
    Task CreateAsync();
    Task ClearAsync();
    Task DeleteAsync();
}

public class AzureServiceBusQueue<T> : IQueue<T>, IQueueManager {
    public async Task EnqueueAsync(/* args */) {
        await EnsureQueueCreatedAsync();

        /* enqueueing stuff */
    }

    private async Task EnsureQueueCreatedAsync() {
        /* check queue exists / lazy init */
        await CreateAsync();
     }

    Task IQueueManager.CreateAsync() {
        /* queue creation stuff */
    }
}
```

moswald:
In that scenario, how would you get an instance of the `IQueueManager` in order to create an `IQueue`?

moswald:
I don't understand the purpose of making the `IQueueManager` implementation explicit. Can you explain?

jamie94bc:
The purpose was to keep the API of `IQueue<T>` simple and without having the additional methods exposed on the class.

I'm tempted to say a completely different architecture/API might work...

```c#
public interface IQueueManager {
    Task<IQueue<T>> CreateQueueAsync<T>(/* opts */);
    Task<bool> QueueExistsAsync(string queueName);
}

public interface IQueue {
    Task ClearAsync();
    Task DeleteAsync();
}
```
With an implementation such as:
```c#
public class AzureServiceBusQueueManager : IQueueManager {
    private readonly NamespaceManager _namespaceManager;

    public AzureServiceBusQueueManager(string connectionString) {
            _namespaceManager = NamespaceManager.CreateFromConnectionString(connectionString);
    }

    public async Task<IQueue<T>> CreateQueueAsync<T>(string queueName) {
        /* creation etc. */

        return new AzureServiceBusQueue(await _namespaceManager.GetQueueAsync(queueName));
    }
}

public class AzureServiceBusQueue : IQueue {
    private readonly QueueClient _queueClient;

    internal AzureServiceBusQueue(QueueClient queueClient) {
        this._queueClient = queueClient;
    }

    public async Task<string> EnqueueAsync(T data) {
        var message = new BrokeredMessage(data);
        await _queueClient.SendAsync(message);
        return message.MessageId;
    }
}
```
and used like...
```c#
container.RegisterSingleton<IQueue<MyQueueItem>>(await new AzureServiceBusQueueManager().CreateQueueAsync("myqueue"));
```

Alternatively, I think both a static async constructor and lazy creation would work too.

Thoughts @niemyjski, @ejsmith?

src/Azure/Queues/AzureServiceBusQueue.cs
ejsmith: I agree about the parameter list being unwieldy. We need to come up with a better creation strategy for sure.

niemyjski: I think by default the manager interface could be implemented in the same queue 
Private task Iqueuemanager.CreateAsync() {}
This would simplify things and not expose too much. We could also use an options class to keep the parameters down and extend it in different ways
Sent from Outlook Mobile

ejsmith:
What about something like this:

```
public interface IQueueManager<TOptions> where TOptions: class {
    Task<IQueue<T>> CreateAsync<T>(TOptions options) where T: class;
    Task<bool> ExistsAsync<T>() where T : class;
    Task<bool> DeleteAsync<T>() where T : class;
}

public class ServiceBusQueueManager : IQueueManager<ServiceBusQueueOptions> {
    public Task<IQueue<T>> CreateAsync<T>(ServiceBusQueueOptions options) where T : class {
        // create queue
        return Task.FromResult<IQueue<T>>(null);
    }

    public Task<bool> ExistsAsync<T>() where T : class {
        return Task.FromResult(false);
    }

    public Task<bool> DeleteAsync<T>() where T : class {
        return Task.FromResult(true);
    }
}

public class ServiceBusQueueOptions {
    public string ConnectionString { get; set; }
    public bool SomeOtherOption { get; set; }
}

public class SampleQueue : IQueue<MyItem> {
    private readonly IQueueManager<ServiceBusQueueOptions> _queueManager;
    private readonly ServiceBusQueueOptions _options;
         
    public SampleQueue(IQueueManager<ServiceBusQueueOptions> queueManager, ServiceBusQueueOptions options) {
        _queueManager = queueManager;
        _options = options;
    }

    private async Task InitializeAsync() {
        if (_initialized)
            return;
       
        if (await _queueManager.ExistsAsync<MyItem>())
            return;

        await _queueManager.CreateAsync<MyItem>(_options);
    }

    public Task<string> EnqueueAsync(MyItem data) {
        await InitializeAsync();
    }

    // rest of implementation
}
```

jamie94bc:
Yeah something like that would work, but the connection string should be on a per manager basis rather than in the options (else exists/delete would require a connection string overload). I think it might be worth splitting the generic interface of `IQueueManager` though and giving options to specify a queue name with/without type param too.

```c#
public interface IQueueManager {
    Task<IQueue<T>> CreateQueueAsync<T>(string queueName = null);
    Task QueueExistsAsync(string queueName);
    Task QueueExistsAsync<T>(string queueName = null);
    Task DeleteQueueAsync(string queueName);
    Task DeleteQueueAsync<T>(string queueName = null);
}

public interface IQueueManager<TOptions> : IQueueManager {
    Task CreateQueueAsync<T>(TOptions opts, string queueName = null);
}
```

Not sure how much this would be used across all implementations (say InMemory) though? We could easily do lazy init without the queue manager interface which would solve #36

moswald:
Abandoning this one in favor of #38.

PR #38, opened 2016-04-15
Lazy queue creation, don't recreate on delete

jamie94bc:
As dicussed in #25, #36, #37 and on Gitter.

Need to do AzureStorageQueue too - RedisQueue already permenently deletes and doesn't have network requests in the ctor, so that should be fine.


jamie94bc:
I think it would be a good idea if `QueueBase` handled calling `EnsureQueueCreatedAsync()`. Couple of ways to do this:

1. Make the abstract methods virtual and update all implementations to call base (if appropriate).

2. Rename the abstract methods to to `{MethodName}Impl`, and implement the behavior in the public method... eg.

```c#
public abstract Task<string> EnqueueAsync(T data);
```
becomes...
```c#
public Task<string> EnqueueAsync(T data) {
    await EnsureQueueCreatedAsync().AnyContext();
    await EnqueueAsyncImpl(data).AnyContext();
}

protected abstract Task<string> EnqueueAsyncImpl(T data);
```

jamie94bc:
Would it be worth accounting for the scenario where a queue might be deleted by another instance of the implementation? I didn't do that for the current implementation primarily because I don't want to be making an additional network request (in the case of Azure Storage / Service Bus) for almost every API call. Consider the below, or in a scale out scenario...

```c#
var a = new AzureServiceBusQueue("myqueue");
var b = new AzureServiceBusQueue("myqueue");

a.EnqueueAsync(new Thing());
var bEntry = b.DequeueAsync();

a.DeleteQueueAsync();
b.EnqueueAsync(new Thing()); // Throws as queue is deleted
a.EnqueueAsync(new Thing()); // Works as the class knows the queue is deleted and recreates
```

We could handle this by adding something like:
```c#
public abstract class QueueBase<T> {
    private bool _queueCreated;

    protected abstract Task EnsureQueueCreatedImplAsync(CancellationToken cancellationToken);
    protected Task EnsureQueueCreatedAsync(CancellationToken cancellationToken = default(CancellationToken)) {
        if (!_queueCreated) {
            await EnsureQueueCreatedImplAsync(cancellationToken).AnyContext();
            _queueCreated = true;
        }
    }

    protected async Task<TResult> WrapQueueCall<TResult>(Func<Task<TResult>> action, CancellationToken cancellationToken) where TResult : Task {
        await EnsureQueueCreatedAsync(cancellationToken);
        try {
            return await action();
        }
        catch (Exception ex) when (IsQueueDeletedException(ex)) {
            _queueCreated = false;
            await EnsureQueueCreatedAsync(cancellationToken);
            return await action();
        }
    }

    protected abstract bool IsQueueDeletedException(Exception ex);

    protected abstract Task<string> EnqueueImplAsync(T data);
    public async Task<string> EnqueueAsync(T data) {
        return await WrapQueueCall(() => EnqueueImplAsync(data));
        await EnsureQueueCreatedAsync().AnyContext();
        return await .AnyContext();
    }
}

public class AzureStorageQueue<T> : QueueBase<T> {
    protected override bool IsQueueDeletedException(Exception ex) => ex is StorageException && ex.Message.Contains("404");
}

**Comment by niemyjski** [comment]
I think if you delete it you should have to create it again

**Comment by jamie94bc** [comment]
Not sure I agree although not a common scenario.

If that is the case then we need a `ClearQueueAsync()` method (#25) and we also need to wrap the exception that is thrown (like Azure Storage SDK wraps WebExceptions with StorageExceptions).

File: src/Azure/Queues/AzureServiceBusQueue.cs [hunk 1]
```diff
@@ -133,40 +149,50 @@ public class AzureServiceBusQueue<T> : QueueBase<T> where T : class {
                     if (!queueEntry.IsAbandoned && !queueEntry.IsCompleted)
                         await queueEntry.AbandonAsync().AnyContext();
                 }
+            }, new OnMessageOptions {
+                AutoComplete = false
             });
```
niemyjski: What does this control? Shouldn't the auto complete setting be set from the constructor?

src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -133,40 +149,50 @@ public class AzureServiceBusQueue<T> : QueueBase<T> where T : class {
...
+            await EnsureQueueCreatedAsync().AnyContext(); // Azure SB needs to call this as it populates the _queueClient field
+
             Interlocked.Increment(ref _abandonedCount);
             await _queueClient.AbandonAsync(new Guid(entry.Id)).AnyContext();
             await OnAbandonedAsync(entry).AnyContext();
         }
         
         public override void Dispose() {
-            _queueClient.Close();
+            _queueClient?.Close();
```
niemyjski: This probably also needs to call dispose (not sure if close does that) and set it to null. Otherwise if you call create again you could be calling a closed or disposed client.

src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -7,73 +7,89 @@
...
-            if (!_namespaceManager.QueueExists(_queueName)) {
-                _queueDescription = new QueueDescription(_queueName) {
-                    MaxDeliveryCount = retries + 1,
-                    LockDuration = _workItemTimeout
-                };
-                _namespaceManager.CreateQueue(_queueDescription);
-            } else {
-                _queueDescription = _namespaceManager.GetQueue(_queueName);
+            using (await _lock.LockAsync(cancellationToken)) {
+                if (_queueClient != null) {
```
niemyjski: This is good, I had a question about this while reviewing the code and wondering what would happen in a multi threaded environment 👍

src/Azure/Queues/AzureServiceBusQueue.cs
jamie94bc: This will tell Service Bus not to autocomplete... Autocompletion is controlled by `queueEntry.CompleteAsync()` and `queueEntry.AbandonAsync()` instead in order to trigger the events.

src/AzureStorage/Queues/AzureStorageQueue.cs
```diff
@@ -32,7 +36,7 @@ public class AzureStorageQueue<T> : QueueBase<T> where T : class {
 
             _queueName = queueName;
             _queueReference = client.GetQueueReference(queueName);
-            _deadletterQueueReference = client.GetQueueReference(String.Concat(queueName, "-deadletter"));
+            _deadletterQueueReference = client.GetQueueReference($"{queueName}-deadletter");
```
niemyjski: Does this do anything network related? If so, I wasn't sure if this should be moved to the ensure method.

src/Redis/Queues/RedisQueue.cs
```diff
@@ -79,6 +79,11 @@ public class RedisQueue<T> : QueueBase<T> where T : class {
             _logger.Trace("Queue {0} created. Retries: {1} Retry Delay: {2}", QueueId, _retries, _retryDelay.ToString());
         }
 
+        protected override Task EnsureQueueCreatedAsync(CancellationToken cancellationToken = new CancellationToken()) {
+            // TODO: Implement RedisQueue.EnsureQueueCreatedAsync
```
niemyjski: Guess we still need to do this :)

src/Redis/Queues/RedisQueue.cs
```diff
@@ -176,7 +181,7 @@ public class RedisQueue<T> : QueueBase<T> where T : class {
             return String.Concat("q:", _queueName, ":in");
         }
 
-        public override async Task<string> EnqueueAsync(T data) {
+        protected override async Task<string> EnqueueImplAsync(T data) {
             string id = Guid.NewGuid().ToString("N");
             _logger.Debug("Queue {_queueName} enqueue item: {id}", _queueName, id);
```
niemyjski: Would be good to move the logging into the base class! and maybe pass id to the imp method? or should we leave id generation up to the client?

@ejsmith @jamie94bc

src/AzureStorage/Queues/AzureStorageQueue.cs
```diff
@@ -153,17 +169,17 @@ public class AzureStorageQueue<T> : QueueBase<T> where T : class {
...
 
-        public override void StartWorking(Func<IQueueEntry<T>, CancellationToken, Task> handler, bool autoComplete = false, CancellationToken cancellationToken = new CancellationToken()) {
+        protected override void StartWorkingImpl(Func<IQueueEntry<T>, CancellationToken, Task> handler, bool autoComplete = false, CancellationToken cancellationToken = new CancellationToken()) {
             if (handler == null)
                 throw new ArgumentNullException(nameof(handler));
-
+            
             var linkedCancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_queueDisposedCancellationTokenSource.Token, cancellationToken).Token;
 
             Task.Run(async () => {
```
niemyjski: In the base implementation we were calling Wait and unwrap exceptions? I'm curious on how that will work I think we had deadlocks in the past because of this. Maybe this code should be moved out to the base class.. I think we might need to have another look at this. Maybe start working should just return a Task and we handle iti n the base implementation? But then the handler passed in returns a Task... Can this be simplified?

@ejsmith

src/Core/Queues/QueueBase.cs
```diff
@@ -38,13 +49,25 @@ public virtual Task<IQueueEntry<T>> DequeueAsync(TimeSpan? timeout = null)
...
+            await EnsureQueueCreatedAsync().AnyContext();
+            return await GetQueueStatsImplAsync().AnyContext();
+        }
 
         public abstract Task DeleteQueueAsync();
         
-        public abstract void StartWorking(Func<IQueueEntry<T>, CancellationToken, Task> handler, bool autoComplete = false, CancellationToken cancellationToken = default(CancellationToken));
+        protected abstract void StartWorkingImpl(Func<IQueueEntry<T>, CancellationToken, Task> handler, bool autoComplete, CancellationToken cancellationToken);
+        public void StartWorking(Func<IQueueEntry<T>, CancellationToken, Task> handler, bool autoComplete = false, CancellationToken cancellationToken = default(CancellationToken)) {
+            EnsureQueueCreatedAsync(cancellationToken).WaitAndUnwrapException(cancellationToken);
```
niemyjski: I think we may need to rework how this is called and what it returns.

src/Azure/Queues/AzureServiceBusQueue.cs
ejsmith: I think Close is good.

src/Redis/Queues/RedisQueue.cs
ejsmith: Yeah, the more code that can be shared, the better.

src/Azure/Queues/AzureServiceBusQueue.cs
jamie94bc: There is no dispose on queue client!

src/AzureStorage/Queues/AzureStorageQueue.cs
jamie94bc: Nope: https://github.com/Azure/azure-storage-net/blob/6152eb5fe3264eaa12117977edcf971b8c3debd2/Lib/Common/Queue/CloudQueueClient.Common.cs#L112

src/AzureStorage/Queues/AzureStorageQueue.cs
jamie94bc: Easiest solution here would be to make `StartWorking` async, @ejsmith ?

ejsmith: I don't think StartWorking really shouldn't be awaited because its non-deterministic and is really meant to just kick off a long running background worker. It has a cancellationtoken that can be used to stop the worker. We already debated about this and decided on this solution, but I am definitely open to discussing it some more.

jamie94bc: We shouldn't get a deadlock with `AnyContext()` so perhaps a non-issue? The other option is to stick it inside a `Task.Run(() => {})` (which some of the implementations already use) but then we wouldn't know when it fails to start working.

niemyjski: I think that the base implementation should just do a Task.Run() and pass it the func and cancellation token.

niemyjski: I wonder if we can make the StartWorkingImp return a task and implify the func that is passed in? I'm sure we did the func that way because it made stuff easier and simplified it a bit.. then the StartWorking can just do taskrun and call the StartWorkingImp.

jamie94bc: What if the call fails though?

jamie94bc: `StartWorkingAsync` is the best option IMO!

src/Azure/Queues/AzureServiceBusQueue.cs
niemyjski: So how would a user delete a queue or set it to null? I guess once it's created ensure is a noop.

src/AzureStorage/Queues/AzureStorageQueue.cs
niemyjski: I don't even see how it's being called.. Looks like only the tests are being called @ejsmith

niemyjski: You might want to know if it fails but I think the queue is self healing so not too worried.

ejsmith: So you want to change it to be Async and then when would the Task complete?  What benefit are we getting from it returning a task?

niemyjski: Well it seems that we are doing async operations in there and then we can have the base operation do a Task.Run and have that await the Task.

jamie94bc: @ejsmith The benefit is we can call `EnsureQueueCreatedAsync` without `.WaitAndUnwrapException()` - `Task.Run` would still be required as a fire and forget (except service bus which has an `OnMessageAsync` callback).

ejsmith: Ahh yeah, I see why you want to do it. But I'm worried it's going to be confusing if it returns a Task. Do you think people will assume that if it returns a Task that it won't complete until it stops listening?

ejsmith: I guess it's reasonable to say that you are awaiting the Task to make sure that you are ready to receive messages. I am good with making this change.

jamie94bc:
Everyone happy to merge? :smile:

niemyjski:
As soon as the build passes :)

jamie94bc:
I think @ejsmith was going to look into that Redis test :wink:

src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -7,73 +7,89 @@
...
 
-                bool changes = false;
+                QueueDescription queueDescription;
 
-                int newMaxDeliveryCount = retries + 1;
-                if (_queueDescription.MaxDeliveryCount != newMaxDeliveryCount) {
-                    _queueDescription.MaxDeliveryCount = newMaxDeliveryCount;
-                    changes = true;
+                if (!await _namespaceManager.QueueExistsAsync(_queueName).AnyContext()) {
+                    queueDescription = await _namespaceManager.CreateQueueAsync(new QueueDescription(_queueName) {
```
moswald: You should catch the exception thrown when `CreateQueueAsync` fails because the queue already exists.

We've seen this in our current project using Foundatio: in a multi-process environment, `QueueExistsAsync` can return `false`, for multiple processes, and then they each call `CreateQueueAsync`, but only one will succeed. The rest will fail with either `Microsoft.ServiceBus.Messaging.MessagingException` or `Microsoft.ServiceBus.Messaging.MessagingEntityAlreadyExistsException`.

The correct solution is to catch those exceptions, and then fall back on calling `GetQueueAsync`. If that _also_ fails, then you can bail. :)

jamie94bc: Hadn't noticed that, I'll get that sorted and see if i can add a test to cover this also.

niemyjski: We could add a distributed lock around this or ball out.

src/Azure/Queues/AzureServiceBusQueue.cs
moswald: Since `DeleteQueueAsync` nulls out the `_queueClient`, I think you've still got a race condition here.

src/Azure/Queues/AzureServiceBusQueue.cs
niemyjski: @moswald feel free to join our public slack channel: https://slack.exceptionless.com

jamie94bc: Yeah I did think that, but that makes queue coupled to a distributed locking provider. Rather not do that.

src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -133,40 +149,50 @@ public class AzureServiceBusQueue<T> : QueueBase<T> where T : class {
                     if (!queueEntry.IsAbandoned && !queueEntry.IsCompleted)
                         await queueEntry.AbandonAsync().AnyContext();
                 }
+            }, new OnMessageOptions {
+                AutoComplete = false
             });
         }
 
         public override async Task<IQueueEntry<T>> DequeueAsync(TimeSpan? timeout = null) {
+            await EnsureQueueCreatedAsync().AnyContext();
```
moswald: And one last thing before I have to go: there's another race condition here (and with all calls to `EnsureQueueCreatedAsync`). If another thread calls `DeleteQueueAsync` immediately after this call returns, the call to `_queueClient.ReceiveAsync` will throw.

I tried to solve this in my previous PR by having `EnsureQueueCreatedAsync` return a reference to the queue, so it at least wouldn't throw `NullReferenceException`, but if the queue is deleted then you'll of course get a different exception related to that.

jamie94bc: We could use the same lock in `DeleteQueueAsync()` as `EnsureQueueCreatedAsync()` to prevent this?

I started this PR with it returning the queue client but moved away to have more code in the base class.

jamie94bc: I started this PR with it returning the queue client but moved away to have more code in the base implementation...

We could wrap `_queueClient` in a property that would throw a more meaningful exception? Even updating `EnsureQueueCreatedAsync()` to return the client wouldn't and adding locking in `DeleteQueueAsync()` wouldn't guarantee 100% it would still exist.

moswald: Yeah, I don't have a great solution - either you get a `NullReferenceException`, or you get a different exception due to the queue not existing even though the client is still valid.

/shrug

Issue #39, opened 2016-04-26
Add support for DefaultMessageTimeToLive and AutoDeleteOnIdle in AzureServiceBusQueue


trbenning:
Obviously there are many more "Azure ServiceBus Queue" specific properties that are unsupported, but these 2 are the ones I need right now. Do you guys have a strategy for how you might support these types of properties moving forward?

I started to do a pull request to set the properties via the constructor, but bolting on properties piecemeal seems like an unsustainable strategy.

jamie94bc:
Up until now we've tried to implement equal functionality on each implementation (eg. storage queue has it's own implementation of deadletter, service bus supports this "natively").

Personally i don't have an issue with having different constructor properties per implementation as long as the exposed implementation (`IQueue<T>`) works the same. If we were to implement this across the board, it would be an option in the constructor (either an additional argument or using a `Options` object if this gets too unwieldy).

niemyjski:
Thanks for the pr!

PR #40, opened 2016-04-27
Support for AutoDeleteOnIdle and DefaultMessageTimeToLive in Azure SB Queues

trbenning:
Resolves issue #39.

### DIFF
--- a/src/Azure/Queues/AzureServiceBusQueue.cs
+++ b/src/Azure/Queues/AzureServiceBusQueue.cs
@@ -67,10 +67,12 @@ namespace Foundatio.Queues {
 
                 if (_queueDescription.AutoDeleteOnIdle != _autoDeleteOnIdle) {
                     _queueDescription.AutoDeleteOnIdle = _autoDeleteOnIdle;
+                    changes = true;
                 }
 
                 if (_queueDescription.DefaultMessageTimeToLive != _defaultMessageTimeToLive) {
                     _queueDescription.DefaultMessageTimeToLive = _defaultMessageTimeToLive;
+                    changes = true;
                 }
 
                 if (changes) {
--- a/src/Azure/Queues/AzureServiceBusQueue.cs
+++ b/src/Azure/Queues/AzureServiceBusQueue.cs
@@ -21,13 +21,21 @@ namespace Foundatio.Queues {
         private long _workerErrorCount;
         private readonly int _retries;
         private readonly TimeSpan _workItemTimeout = TimeSpan.FromMinutes(5);
+        private readonly TimeSpan _autoDeleteOnIdle = TimeSpan.MaxValue;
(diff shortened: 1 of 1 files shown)


ejsmith:
This looks good to me.

src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -21,22 +21,33 @@ public class AzureServiceBusQueue<T> : QueueBase<T> where T : class {
...
+        public AzureServiceBusQueue(string connectionString, string queueName = null, int retries = 2, TimeSpan? workItemTimeout = null, bool shouldRecreate = false, RetryPolicy retryPolicy = null, ISerializer serializer = null, IEnumerable<IQueueBehavior<T>> behaviors = null, ILoggerFactory loggerFactory = null, TimeSpan? autoDeleteOnIdle = null, TimeSpan? defaultMessageTimeToLive = null) : base(serializer, behaviors, loggerFactory) {
             _queueName = queueName ?? typeof(T).Name;
             _namespaceManager = NamespaceManager.CreateFromConnectionString(connectionString);
             _retries = retries;
             if (workItemTimeout.HasValue && workItemTimeout.Value < TimeSpan.FromMinutes(5))
                 _workItemTimeout = workItemTimeout.Value;
 
+            if (autoDeleteOnIdle.HasValue && autoDeleteOnIdle.Value >= TimeSpan.FromMinutes(5))
+                _autoDeleteOnIdle = autoDeleteOnIdle.Value;
```
niemyjski: Any reason why we'd want to limit this to timespans greater than 5 minutes? What happens if your app is idle shortly after startup?

jamie94bc: That might be a requirement of Azure like `workItemTimeout.Value < TimeSpan.FromMinutes(5)` @trbenning?

trbenning: @jamie94bc  is right. The minimum value for AutoDeleteOnIdle is 5 minutes.

src/Azure/Queues/AzureServiceBusQueue.cs
```diff
@@ -54,6 +65,14 @@ public class AzureServiceBusQueue<T> : QueueBase<T> where T : class {
                     changes = true;
                 }
 
+                if (_queueDescription.AutoDeleteOnIdle != _autoDeleteOnIdle) {
+                    _queueDescription.AutoDeleteOnIdle = _autoDeleteOnIdle;
+                }
```
niemyjski: @trbenning I think we are missing the changes = true check as in the lines above. Can you please confirm this.

trbenning: Good catch. I've updated the source branch.

niemyjski:
Thanks!

PR #44, opened 2016-05-17
initial changes to adopt project.json programming model.

Bartmax:
This is the first step to just get it onto the new stuff (project.json programming model) using net451.
Next obvious step will be to use netstandard.

What has been done?

Folder name changes. On new world, projects compile into packages, and the folder is the package name. Having packages with same names as external packages doesn't work.

Project inside projects (tests) doesn't look to work either (because same reason as before).
Moved to the new *standard* location `./tests` and using proper folder name to avoid issues.

Some (very few) utility methods were inaccesible (not sure why) and I had to change them from internal to public.

package.config renamed into -> project.json and updated the format for dependencies.

DELETED all .csproj and let VS create the corresponding .xproj

created a new .sln with the corresponding folder structures


niemyjski:
Wonder what all build changes we'll have to make. I think we'll have to update appveyor to install dotnet:

https://github.com/enricosada/fsharp-dotnet-cli-samples/blob/master/appveyor.yml#L15-L20

I'm going to reach out to appveyor right now and ask

niemyjski:
https://github.com/JimBobSquarePants/ImageProcessor/blob/feature/core-v1/appveyor.yml

niemyjski:
http://help.appveyor.com/discussions/questions/2249-how-to-use-latest-dotnet-cli

Think it might be best to do a hybrid of the two installs.

src/Foundatio.Core/Utility/Run.cs
```diff
@@ -6,7 +6,7 @@
 using Foundatio.Logging;
 
 namespace Foundatio.Utility {
-    internal static class Run {
+    public static class Run {
```
niemyjski: This was internal and I believe a linked class that was compiled internally.

src/Foundatio.Core/Utility/TaskHelper.cs
```diff
@@ -3,7 +3,7 @@
 using System.Threading.Tasks;
 
 namespace Foundatio.Utility {
-    internal static class TaskHelper {
+    public static class TaskHelper {
```
niemyjski: This was internal and I believe a linked class that was compiled internally.

src/Foundatio.Redis/Queues/RedisQueue.cs
```diff
@@ -492,7 +492,7 @@ public class RedisQueue<T> : QueueBase<T> where T : class {
                 _monitor.Pulse();
         }
 
-        internal async Task DoMaintenanceWorkAsync() {
+        public async Task DoMaintenanceWorkAsync() {
```
niemyjski: I think we need to figure out how to make this internal.

niemyjski:
https://github.com/feinoujc/Mandrill.net/pull/50/files

src/Foundatio.Core/Utility/TaskHelper.cs
niemyjski: Looks like we can do a file link like this (example): “buildOptions:” { “include”: [“../Folder/**.*.cs”] }

niemyjski:
https://github.com/aspnet/Home/issues/1451

niemyjski:
Unfortunately I have to close this as files were changed in the root making this very hard to merge :(. Thank you for your help

Issue #45, opened 2016-06-10
Race condition in AzureServiceBusQueue when multiple processes are creating the same queue


moswald:
Right now, `EnsureQueueCreatedAsync` will check if the queue exists, and then if not, it will try to create it. If another process creates the queue in-between these two calls, then the create fails with a `MessagingEntityAlreadyExistsException`. This is caught and handled and works fine.

However, there's a catch - if another process is _currently creating_ the queue, then it will fail with `MessagingException`, which is _not_ caught.

The error seen in logs is:

    Microsoft.ServiceBus.Messaging.MessagingException: The remote server returned an error: (409) Conflict. SubCode=40901. Another conflicting operation is in progress.

Since `MessagingEntityAlreadyExistsException` derives from `MessagingException`, the solution is to catch `MessagingException` instead.

moswald:
Fixed by #46

PR #46, opened 2016-06-10
Addresses #45 by catching the base exception

moswald:


ejsmith:
Looks good. Thank you!

PR #49, opened 2016-06-20
adding sets to cache functionality

ttugrad01:


src/Core/Caching/InMemoryCacheClient.cs
```diff
@@ -168,6 +168,8 @@ public class InMemoryCacheClient : MaintenanceBase, ICacheClient {
             if (expiresIn.HasValue)
                 ScheduleNextMaintenance(expiresAt);
 
+            Cleanup();
```
niemyjski: Any reason for doing cleanup here instead of having maintenance do it?

src/Core/Caching/InMemoryCacheClient.cs
```diff
@@ -203,9 +205,84 @@ public class InMemoryCacheClient : MaintenanceBase, ICacheClient {
...
+                return false;
+            }
+
+
+            var entry = new CacheEntry(new List<T> { value }, expiresAt, ShouldCloneValues);
+            _memory.AddOrUpdate(key, entry, (k, cacheEntry) =>
+            {
+                var collection = cacheEntry.Value as ICollection<T>;
+                collection?.Add(value);
+                cacheEntry.Value = collection;
```
niemyjski: @ejsmith @ttugrad01 we need to think about what we should do if the cache value is not an ICollection. This code will set the cache value to null.

src/Core/Caching/InMemoryCacheClient.cs
```diff
@@ -203,9 +205,84 @@ public class InMemoryCacheClient : MaintenanceBase, ICacheClient {
...
+                _logger.Trace("Removing expired key {0}", key);
+
+                await this.RemoveAsync(key).AnyContext();
+                await OnItemExpiredAsync(key).AnyContext();
+                return false;
+            }
+            
+            _memory.TryUpdate(key, (k, cacheEntry) => {
+                var collection = cacheEntry.Value as ICollection<T>;
+                if (collection != null && collection.Contains(value)) {
```
niemyjski: Do we need to do null checks for value? I don't think you can have a null key. Probably should add tests for this if we don't have any.

src/Core/Caching/InMemoryCacheClient.cs
ejsmith: It always has to be a collection. It's a set. We control the underlying storage when people add / remove we put a collection in the cache.

src/Core/Extensions/ConcurrentDictionaryExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+                throw new ArgumentNullException(nameof(updateValueFactory));
+            TValue comparisonValue;
+            TValue newValue;
+            do {
+                if (!concurrentDictionary.TryGetValue(key, out comparisonValue)) {
+                    return false;
+                }
+                newValue = updateValueFactory(key, comparisonValue);
+            }
+            while (!concurrentDictionary.TryUpdate(key, newValue, comparisonValue));
```
niemyjski: should this bail out after so many attempts? @ejsmith thoughts?

src/Core/Caching/InMemoryCacheClient.cs
ejsmith: I guess if someone else set a normal value in this key it might be a problem. Maybe we should bail throw or just bail and return false if it's not a collection.

src/Core/Caching/InMemoryCacheClient.cs
ejsmith: Yeah, probably need to check the key for null or empty.

src/Core/Extensions/ConcurrentDictionaryExtensions.cs
ejsmith: This is the same code stolen from AddOrUpdate in the .NET framework.

niemyjski:
@ejsmith @ttugrad01 looks good. I almost wonder if this should be declared as a separate interface. Adding this directly onto the core cache client makes it much harder to implement a new cache client especially when this functionality may not be supported.

ejsmith:
We thought about that and this is a pretty common use case and it's not that hard to implement. I think it's good being in this contract.

ejsmith:
Changes look good to me.

src/Redis/Tests/Caching/RedisCacheClientTests.cs
```diff
@@ -68,8 +69,19 @@ public class RedisCacheClientTests : CacheClientTestsBase {
...
-            return base.CanManageSets();
+        public override async Task CanManageSets() {
+            await base.CanManageSets();
+
+            var cache = GetCacheClient();
+            if (cache == null)
+                return;
+
+            using (cache) {
+                await Assert.ThrowsAsync<StackExchange.Redis.RedisServerException>(async () => {
```
niemyjski: I believe we should be checking this in all the cache clients the same way. And verify the key value hasn't been changed or set to null (do we agree on this @ejsmith).

ejsmith: Yeah, we talked about that, but the problem is that redis throws generic exceptions so we can't really handle that error specifically.

PR #52, opened 2016-07-15
Rabbit mq

barkhashah:
There are 4 applications for review:
1) Foundatio.RabbitMQ - This is the main library that is implementing pub/sub model of RabbitMQ message broker.
2) Foundatio.RabbitMQ.Tests - Unit tests written for Foundatio.RabbitMQ
3) Foundatio.RabbitMQSubscribeConsole - Sample application for subscriber that uses Foundatio.RabbitMQ 
5) Foundatio.RabbitMQPublishConsole - Sample application for publisher that uses Foundatio.RabbitMQ


niemyjski:
@moswald @trbenning Thank you a ton for submitting a PR. I'll try and review this soon. Quick question, is there any advantage to targeting rabbitMq directly rather than using AMQP protocol?

On a side note It's kind of hard to review it since we moved master to feature/dotnet branch and now it says a ton of things have changed... We did this so we can wait for some of our dependencies to stabilize a bit more 😵.

niemyjski:
Also, we welcome you all to join our slack and hang out! http://slack.exceptionless.com

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageService.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <summary>
+        /// Subscribe for the message
+        /// </summary>
+        /// <typeparam name="T">Type of the subscriber who wants to be notified of the callback</typeparam>
+        /// <param name="handler">callback handler</param>
+        /// <param name="cancellationToken"></param>
+        public override void Subscribe<T>(Func<T, CancellationToken, Task> handler,
+            CancellationToken cancellationToken = default(CancellationToken)) {
+            var consumer = new EventingBasicConsumer(_subscriberChannel);
+            consumer.Received += async (o, e) => {
```
niemyjski: @trbenning @moswald should we be worried about not being able to unsubscribe on dispose?

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageService.cs
```diff
@@ -123,17 +139,23 @@ public class RabbitMQMessageService : MessageBusBase, IMessageBus {
...
+        {
+            var data = await _serializer.SerializeAsync(new MessageBusData
+            {
                 Type = messageType.AssemblyQualifiedName,
                 Data = await _serializer.SerializeToStringAsync(message).AnyContext()
             }).AnyContext();
-            // The publication occurs with mandatory=false
-            _publisherChannel.BasicPublish(_exchangeName, _routingKey, basicProperties, data);
+
+            lock (_lock) {
```
niemyjski: Bit worried about this right here... In the async context you can use AsyncLock which provides .Lock() and .LockAsync() methods that are awaitable.

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageService.cs
```diff
@@ -178,7 +198,11 @@ public class RabbitMQMessageService : MessageBusBase, IMessageBus {
             if (_subscriberClient != null && _subscriberClient.IsOpen) {
                 _subscriberClient.Close();
             }
-            _publisherChannel?.Abort();
+            if (_lock != null) {
+                lock (_lock) {
+                    _publisherChannel?.Abort();
+                }
+            }
             _subscriberChannel?.Abort();
```
niemyjski: Any reason for locking around one bot not the other? Just curious 👍

niemyjski:
@moswald @trbenning Looks great! I just had a few questions above that may prevent you from hitting dead lock scenarios :). I'll start looking for a way to merge this into the other branch.

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageService.cs
barkhashah: @niemyjski If the object gets disposed then there is no need to worry about unsubscription. I have tested it myself where once the Dispose is called where we are closing the connection and channels, the event handler never gets called.

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageService.cs
barkhashah: @niemyjski  In the RabbitMQ documentation, it clearly says that IModel instances should not be used by more than one thread simoultaneously. Logically publish message is the one that can be used across mulitple threads, however, when it comes to subscription we do that only once . So couldnt think of any scenario where we will call subscribe many times from the same or multiple threads. If its across process then ofcourse the RabbitMQMessageService intance will be different.  Atleast that was my thinking :)

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageService.cs
barkhashah: @niemyjski Thanks for pointing it out. Will make a change and use AsyncLock from Nito.AsyncEx

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageService.cs
niemyjski: We may want to change this up, because we should always clean up after ourself to prevent leaking memory (queues should be long lived, but we don't know how the consumer is going to use it. Do you agree?

niemyjski:
I'm closing this in favor of the other pr :)

PR #53, opened 2016-07-15
Rabbit mq

barkhashah:
Merging RabbitMQ library, sample application and unit tests in the feature/dotnet

### DIFF
--- a/Foundatio.sln
+++ b/Foundatio.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{70515E66-DAF8-4D18-8F8F-8A2934171AA9}"
 EndProject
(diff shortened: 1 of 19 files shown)


niemyjski:
I'm going to merge this in, we may or may not keep the two sample applications. We'll have to think about it. Everything else looks good :) Thank you!!!!!!!!

niemyjski:
Thank you for your contribution. It's greatly appreciated! Do you know if there is a good way to setup rabbit mq on appveyor so we can get the tests running?

niemyjski:
I just found this https://gist.github.com/FeodorFitsner/06b377df42ba8f398909 looks like we just need to put that script in the tools folder and call it next to our appveyor script for installing redis.. Would you mind doing that?

niemyjski:
I'll merge this in and install the script, will only take me a minute :)

niemyjski:
Everything has been merged, tested and deployed Thanks a bunch! I had one last question...


```

        private void CloseConnection() {
            if (_publisherClient != null && _publisherClient.IsOpen)
                _publisherClient.Close();
            if (_subscriberClient != null && _subscriberClient.IsOpen)
                _subscriberClient.Close();

            if (_lock != null) {
                using (_lock.Lock())
                    _publisherChannel?.Abort();
            }

            _subscriberChannel?.Abort();
        }
```

Any reason for locking publisher channel? Why not the subscriber channel. I would of thought channels would need to be closed before clients... Is this correct? You are constructing the channels from the client but disposing of them the opposite way.

barkhashah:
@niemyjski - Actually as per the RabbitMQ documentation, channel should not be shared across the threads. I only locked publisher channel because publishing messages from the multiple threads of the same process is very likely compared to subscribing the event on every thread. We subscribe only once , however, we can publish as many times on the same process. Please let me know if I am wrong in thinking here or suggest me if you see any other possibility where we can subscribe on the same channel on different threads.

>> I would of thought channels would need to be closed before clients... Is this correct?
I agree. Channel should be closed first and then client connections :+1: 
I will go ahead and make this change.

ejsmith:
@barkhashah also, I noticed that Rabbit supports delayed messages. So we should use that instead of the timer based in memory delay since that has the potential to lose messages.

PR #55, opened 2016-07-21
Improved job logging

niemyjski:


src/Foundatio/Jobs/IJob.cs
```diff
@@ -40,32 +40,29 @@ public static class JobExtensions {
                         else if (iterations % 1000 == 0) // allow for cancellation token to get set
                             await SystemClock.SleepAsync(1).AnyContext();
                     } catch (TaskCanceledException) { }
-
-                    if (continuationCallback == null)
+                    
+                    if (continuationCallback == null || cancellationToken.IsCancellationRequested)
```
niemyjski: This saves us from running the callback if cancellation is requested.

src/Foundatio/Jobs/IJob.cs
```diff
@@ -40,32 +40,29 @@ public static class JobExtensions {
...
                 }
-
-                logger.Info("Stopping continuous job type \"{0}\" on machine \"{1}\"...", jobName, Environment.MachineName);
-
+                
                 if (cancellationToken.IsCancellationRequested)
                     logger.Trace("Job cancellation requested.");
 
-                await SystemClock.SleepAsync(1).AnyContext(); // allow events to process
+                logger.Info("Stopping continuous job type \"{0}\" on machine \"{1}\"...", jobName, Environment.MachineName);
```
niemyjski: It doesn't make any sense to sleep after a continuous run is completed. If anything this should be inside of the while loop

src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
```diff
@@ -39,65 +39,58 @@ public class WorkItemJob : IQueueJob, IHaveLogger {
...
             }
 
-            Type workItemDataType = null;
+            Type workItemDataType;
             try {
                 workItemDataType = Type.GetType(queueEntry.Value.Type);
             } catch (Exception ex) {
                 await queueEntry.AbandonAsync().AnyContext();
-                return JobResult.FromException(ex, $"Could not resolve {workItemDataType.Name} work item data type.");
+                return JobResult.FromException(ex, $"Abandoning {queueEntry.Value.Type} work item: {queueEntry.Id}: Could not resolve work item data type.");
```
niemyjski: This and the below code could have thrown a null reference exception..

src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
```diff
@@ -39,65 +39,58 @@ public class WorkItemJob : IQueueJob, IHaveLogger {
...
+                return JobResult.FailedWithMessage($"Completing {queueEntry.Value.Type} work item: {queueEntry.Id}: Handler for type {workItemDataType.Name} not registered.");
             }
 
             if (queueEntry.Value.SendProgressReports)
-                await _messageBus.PublishAsync(new WorkItemStatus {
-                    WorkItemId = queueEntry.Value.WorkItemId,
-                    Progress = 0,
-                    Type = queueEntry.Value.Type
-                }).AnyContext();
+                await ReportProgress(handler, queueEntry).AnyContext();
```
niemyjski: I moved this to a method to remove some repetitive ness but also wrapped it in a try catch with logging. Reporting progress should never blow up job processing.

The question is, should these be a run with retries?

src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
```diff
@@ -39,65 +39,58 @@ public class WorkItemJob : IQueueJob, IHaveLogger {
...
-
-                if (handler.AutoRenewLockOnProgress)
-                    await lockValue.RenewAsync().AnyContext();
+                if (handler.AutoRenewLockOnProgress) {
+                    try {
+                        await queueEntry.RenewLockAsync().AnyContext();
+                        await lockValue.RenewAsync().AnyContext();
+                    } catch (Exception ex) {
+                        handler.Log.Error(ex, "Error renewing work item locks: {0}", ex.Message);
+                    }
```
niemyjski: Should renewing locks be on a retry as well? I'm wondering if these should be passed a cancellation token?

niemyjski:
@ejsmith one thing to also considered is the log messages are only written, if abandon or complete don't throw an exception. Which it could for what ever reason..

ejsmith:
Looks good

PR #56, opened 2016-07-26
rabbitmq delayed message exchange

barkhashah:
This change implements rabbitmq delayed message exchange. You will be required to enable the plugin rabbitmq_delayed_message_exchange. Added a member in the constructor that decides if the user wants to create and actual exchange or the delayed exchange.

### DIFF
--- a/src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageBus.cs
+++ b/src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageBus.cs
@@ -13,15 +13,11 @@ using RabbitMQ.Client.Exceptions;
 namespace Foundatio.Messaging {
     public class RabbitMQMessageBus : MessageBusBase, IMessageBus {
         private readonly string _queueName;
-        private readonly string _routingKey;
+        private readonly string _routingKey = "";
         private readonly string _exchangeName;
-        private readonly bool _durable;
-        private readonly bool _persistent;
-        private readonly bool _exclusive;
-        private readonly bool _autoDelete;
         private readonly IDictionary<string, object> _queueArguments;
         private readonly ISerializer _serializer;
-        private readonly IConnectionFactory _factory;
+        private readonly ConnectionFactory _factory;
         private IConnection _publisherClient;
         private IConnection _subscriberClient;
         private IModel _publisherChannel;
@@ -30,40 +26,36 @@ namespace Foundatio.Messaging {
         private readonly TimeSpan _defaultMessageTimeToLive = TimeSpan.MaxValue;
 
         /// <summary>
-        /// Constructor for RabbitMqMessaging - Exchange type set as Direct exchange that uses the routing key
+        /// Constructor for RabbitMqMessaging - Exchange type set as fanout exchange that uses the exchange name
         /// </summary>
         /// <param name="userName">username needed to create the connection with the broker</param>
         /// <param name="password">password needed to create the connection with the broker</param>
-        /// <param name="queueName">Queue name</param>
-        /// <param name="routingKey">The routing key is an "address" that the exchange may use to decide how to route the message</param>
         /// <param name="exhangeName">Name of the direct exchange that delivers messages to queues based on a message routing key</param>
-        /// <param name="durable">Durable exchanges survive broker restart</param>
-        /// <param name="autoDelete">True, if you want the queue to be deleted when the connection is closed</param>
+        /// <param name="queueName">Name of the queue established by the subscriber when they call QueueDeclare. Its not used by publisher.</param>
         /// <param name="queueArguments">queue arguments</param>
         /// <param name="defaultMessageTimeToLive">The value of the expiration field describes the TTL period in milliseconds</param>
         /// <param name="serializer">For data serialization</param>
         /// <param name="loggerFactory">logger</param>
-        /// <param name="persistent">When set to true, RabbitMQ will persist message to disk</param>
-        /// <param name="exclusive"></param>
-        public RabbitMQMessageBus(string userName, string password, string queueName, string routingKey,
-            string exhangeName, bool durable, bool persistent, bool exclusive, bool autoDelete,
+        /// <remarks>https://www.rabbitmq.com/dotnet-api-guide.html#connection-recovery</remarks>
+        public RabbitMQMessageBus(string userName, string password, string exhangeName, string queueName,
             IDictionary<string, object> queueArguments = null, TimeSpan? defaultMessageTimeToLive = null,
             ISerializer serializer = null, ILoggerFactory loggerFactory = null) : base(loggerFactory) {
             _serializer = serializer ?? new JsonNetSerializer();
             _exchangeName = exhangeName;
             _queueName = queueName;
-            _routingKey = routingKey;
-            _durable = durable;
-            _persistent = persistent;
-            _exclusive = exclusive;
-            _autoDelete = autoDelete;
             _queueArguments = queueArguments;
             if (defaultMessageTimeToLive.HasValue && defaultMessageTimeToLive.Value > TimeSpan.Zero)
                 _defaultMessageTimeToLive = defaultMessageTimeToLive.Value;
-            // initialize the connection factory
-            _factory = new ConnectionFactory {
+            // initialize the connection factory. automatic recovery will allow the connections to be restored
+            // in case the server is restarted or there has been any network failures
+            // Topology ( queues, exchanges, bindings and consumers) recovery "TopologyRecoveryEnabled" is already enabled
+            // by default so no need to initialize it. NetworkRecoveryInterval is also by default set to 5 seconds.
+            // it can always be fine tuned if needed.
+            _factory = new ConnectionFactory
+            {
                 UserName = userName,
-                Password = password
+                Password = password,
+                AutomaticRecoveryEnabled = true
             };
             // initialize the publisher
             InitPublisher();
@@ -102,7 +94,6 @@ namespace Foundatio.Messaging {
             }
 
             var basicProperties = _publisherChannel.CreateBasicProperties();
-            basicProperties.Persistent = _persistent;
             basicProperties.Expiration = _defaultMessageTimeToLive.TotalMilliseconds.ToString(CultureInfo.InvariantCulture);
             // RabbitMQ only supports delayed messages with a third party plugin called "rabbitmq_delayed_message_exchange"
             if (_delayedExchangePluginEnabled && delay.HasValue && delay.Value > TimeSpan.Zero) {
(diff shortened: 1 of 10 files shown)


src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageBus.cs
```diff
@@ -117,21 +131,22 @@ public class RabbitMQMessageBus : MessageBusBase, IMessageBus {
...
                 Data = await _serializer.SerializeToStringAsync(message).AnyContext()
             }).AnyContext();
             
             var basicProperties = _publisherChannel.CreateBasicProperties();
             basicProperties.Persistent = _persistent;
-            basicProperties.Expiration = _defaultMessageTimeToLive.Milliseconds.ToString();
-
+            basicProperties.Expiration = _defaultMessageTimeToLive.TotalMilliseconds.ToString(CultureInfo.InvariantCulture);
+            // RabbitMQ only supports delayed messages with a third party plugin called "rabbitmq_delayed_message_exchange"
+            if (_delayedExchange && delay.HasValue && delay.Value > TimeSpan.Zero) {
```
niemyjski: Is there any way to tell if this plugin is installed a head of time and throw an exception?? Or what will happen here if the plugin is not installed?

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageBus.cs
```diff
@@ -117,21 +131,22 @@ public class RabbitMQMessageBus : MessageBusBase, IMessageBus {
                 return;
             _logger.Trace("Message Publish: {messageType}", messageType.FullName);
 
-            // RabbitMQ only supports delayed messages with a third party plugin.
-            if (delay.HasValue && delay.Value > TimeSpan.Zero) {
```
niemyjski: I think this should still happen if you don't have _delayExchange set to true.. This will at least give you some kind of basic delayed message (as long as your app doesn't crash).

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageBus.cs
```diff
@@ -105,7 +119,7 @@ public class RabbitMQMessageBus : MessageBusBase, IMessageBus {
         /// </summary>
         /// <param name="messageType"></param>
         /// <param name="message"></param>
-        /// <param name="delay"></param>
+        /// <param name="delay">Along with the delay value, _delayExchange should also be set to true</param>
```
niemyjski: This shouldn't be required as we do have a base class to do delayed messages (in process) while not as good or reliable, this is an implementation detail :)

niemyjski:
Thanks for the commit, I reviewed it and left a few comments. I think we should fall back to using the base class for delayed messaging.

Also, if you want to merge into master instead of the .net branch your changes will go live sooner. There was a time period where dotnet branch was master and we changed it around and merged in all the non .net core changes into master.

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageBus.cs
barkhashah: Well I tried to look for an api to find out if there is a way to know if plugin is installed but couldn't find one. But yes, if the plugin is not installed and you are trying to create x-delayed-message exchange then the exception gets thrown.

Unhandled Exception: RabbitMQ.Client.Exceptions.OperationInterruptedException: T
he AMQP operation was interrupted: AMQP close-reason, initiated by Peer, code=50
3, text="COMMAND_INVALID - invalid exchange type 'x-delayed-message'", classId=4
0, methodId=10, cause=
   at RabbitMQ.Client.Impl.SimpleBlockingRpcContinuation.GetReply(TimeSpan timeo
ut)
   at RabbitMQ.Client.Impl.ModelBase.ModelRpc(MethodBase method, ContentHeaderBa
se header, Byte[] body)
   at RabbitMQ.Client.Framing.Impl.Model._Private_ExchangeDeclare(String exchang
e, String type, Boolean passive, Boolean durable, Boolean autoDelete, Boolean in
ternal, Boolean nowait, IDictionary`2 arguments)
   at RabbitMQ.Client.Impl.ModelBase.ExchangeDeclare(String exchange, String typ
e, Boolean durable, Boolean autoDelete, IDictionary`2 arguments)
   at Foundatio.Messaging.RabbitMQMessageBus.SetUpExchangeAndQueuesForRouting(IM
odel model) in C:\Users\barkha.shah\Foundatio\src\Foundatio.RabbitMQ\Messaging\R
abbitMQMessageBus.cs:line 98
   at Foundatio.Messaging.RabbitMQMessageBus..ctor(String userName, String passw
ord, String queueName, String routingKey, String exhangeName, Boolean delayedExc
hange, Boolean durable, Boolean persistent, Boolean exclusive, Boolean autoDelet
e, IDictionary`2 queueArguments, Nullable`1 defaultMessageTimeToLive, ISerialize
r serializer, ILoggerFactory loggerFactory) in C:\Users\barkha.shah\Foundatio\sr
c\Foundatio.RabbitMQ\Messaging\RabbitMQMessageBus.cs:line 72
   at Foundatio.RabbitMQSubscribeConsole.Program.Main(String[] args)
Press any key to continue . . .

niemyjski: We could just revert the changes you made to the constructor and the first time a delayed message is called we try to call it and if we get this exception with that specific code, we could just fall back to using the base class behavior for delayed messages. This way it's not an extra option, it just works if configured.

niemyjski: When this exception occurs we could store it in a flag, and then short circuit it if not available so it only happens once.

barkhashah: I introduced a new constructor parameter called "bool delayedExchange" that tells us whether to create delayed exchange or not. Now declaring the exchange happens in the constructor. This variable was the only way that was telling me what kind of exchange to declare ( delayed or actual ) . Now am I understanding this right that you want me to take out "bool delayedExchange" from the constructor and always try to create the delayed exchange and if that fails then create the actual and store the ( exception occured flag for short circuiting future exchange declare as long as its not a new instance of RabbitMQMessageBus) . We come to know only later during publish method that user intends to delay sending the message, however, we create exchange in the constructor itself.

niemyjski: Yes, What do people normally do? You'd think that I'd always want a delayed exchange as it gives me both non delayed and delayed messaging support (right?). If we can remove an option and have the best of both worlds (either delayed messages via in process support or offloaded onto rabbitmq) that would be awesome :)

barkhashah: I agree. However, people need to be smart enough to not use the delayed exchange for sending the non-delayed message. There is a peformance issue with that.

niemyjski: Looked at that link you post, any disadvantages to having two exchanges defined with the same exact name? If we fail to create or publish to the one we dispose of it and set it to null and call one for no delay, and the other or base implementation for delayed messages?

By doing one or the other right now and taking a perf hit kinda sucks, especially when it might be total random when you want to do a delayed message...

barkhashah: Yeah there is a problem if you try to create delayed and then actual exchange or vice versa with the *same name*.

Lets say you start with creating an exchange name as "foundatioExchange" and there is no plugin enabled for delay exchange. Now as per our logic we will always attempt to create x-delayed-type exchange and if that fails we fall back to the actual one. So now we have a successful creation of a regular exchange called "foundatioExchange".

Now you enable the plugin.

After enabling the plugin if you try to create a new RabbitMQMessageBus pub/sub with the same name as "foundatioExchange" you will be thrown with OperationInterruptedException .

+		o	{"The AMQP operation was interrupted: AMQP close-reason, initiated by Peer, code=406, text=\"PRECONDITION_FAILED - inequivalent arg 'type' for exchange 'FoundatioExchange' in vhost '/': received ''x-delayed-message'' but current is 'direct'\", classId=40, methodId=10, cause="}	RabbitMQ.Client.Exceptions.OperationInterruptedException


So you can't keep the same names for 2 different types of exchange. 
I have made some changes where we remove the extra boolean from the constructor and always create the delayed exchange first. If we fail then we fall back to the regular one. Currently testing various schenarios at this point. I will submit for new changes tomorrow.

src/Foundatio.RabbitMQ/Messaging/RabbitMQMessageBus.cs
barkhashah: Well this is the only unit test case which is failing for me. I am still trying to understand its behavior.

Test Name:	CanSendDelayedMessage
Test FullName:	Foundatio.RabbitMQ.Tests.Messaging.RabbitMqMessageBusTests.CanSendDelayedMessage
Test Source:	C:\Users\barkha.shah\Foundatio\test\Foundatio.RabbitMQ.Tests\Messaging\RabbitMqMessageBusTests.cs : line 39
Test Outcome:	Failed
Test Duration:	0:10:34.256

Result StackTrace:	
at Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer)
   at Foundatio.Tests.Messaging.MessageBusTestBase.<CanSendDelayedMessage>d__5.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Xunit.Sdk.TestInvoker`1.<>c__DisplayClass46_1.<<InvokeTestMethodAsync>b__1>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Xunit.Sdk.ExecutionTimer.<AggregateAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Xunit.Sdk.ExceptionAggregator.<RunAsync>d__9.MoveNext()
Result Message:	
Assert.Equal() Failure
Expected: 0
Actual:   998

barkhashah: So when I debug this particular method, I get failure. However, when I run the test seperately it passes.

barkhashah: I have this logic in the PublishAsync now :

`            // if the rabbitmq plugin is not availaible then use the base class delay mechanism
            if (!_delayedExchangePluginEnabled  && delay.HasValue && delay.Value > TimeSpan.Zero) {
                _logger.Trace("Schedule delayed message: {messageType} ({delay}ms)", messageType.FullName, delay.Value.TotalMilliseconds);
                await AddDelayedMessageAsync(messageType, message, delay.Value).AnyContext();
                return;
            }`
When I run the unit test case with the plugin enabled then the test case for CanSendDelayedMessage fails because the above logic will not get executed. With plugin disabled the test case passes.

PR #61, opened 2016-09-01
changing interface for cache sets to multiple allow multiple values; …

ttugrad01:
… fixing issue with redis hybrid cache and getting set data


ejsmith:
LGTM

PR #62, opened 2016-09-15
Rx playground

ejsmith:


src/Foundatio.Logging.Xunit/TestLoggerFactory.cs
```diff
@@ -12,7 +12,7 @@ public class TestLoggerFactory : ILoggerFactory {
             _testOutputHelper = output;
         }
 
-        public LogLevel MinimumLevel { get; set; } = LogLevel.Information;
+        public LogLevel MinimumLevel { get; set; } = LogLevel.Trace;
```
niemyjski: I think this meant to be a temporary change.

src/Foundatio/Jobs/IJob.cs
```diff
@@ -38,8 +38,8 @@ public static class JobExtensions {
                         else if (interval.HasValue)
                             await SystemClock.SleepAsync(interval.Value, cancellationToken).AnyContext();
                         else if (iterations % 1000 == 0) // allow for cancellation token to get set
-                            SystemClock.Sleep(1);
-                    } catch (TaskCanceledException) { }
+                            await SystemClock.SleepAsync(1);
```
niemyjski: Missing AnyContext()

src/Foundatio/Metrics/MetricTimer.cs
```diff
@@ -1,29 +1,29 @@
...
         }
 
         public async Task DisposeAsync() {
             if (_disposed)
                 return;
 
             _disposed = true;
-            _stopWatch.Stop();
-            await _client.TimerAsync(_name, (int)_stopWatch.ElapsedMilliseconds).AnyContext();
+            await _client.TimerAsync(_name, (int)_stopWatch.Elapsed.TotalMilliseconds).AnyContext();
```
niemyjski: Won't the stopwatch keep running or do we need to call dispose on it? I don't know how it's implemented behind the scenes.

src/Foundatio/Utility/ScheduledTimer.cs
```diff
@@ -61,7 +63,8 @@ public class ScheduledTimer : IDisposable {
 
                 _logger.Trace(() => $"Scheduling next: delay={delay}");
 
-                _timer.Change(delay, Timeout.Infinite);
+                DisposeTimer();
+                _timer = SystemClock.Instance.Schedule(TimeSpan.FromMilliseconds(delay), async () => await RunCallbackAsync());
```
niemyjski: What about a race condition between when schedule next is called a bunch? I think that's one of the reasons we were calling change on the timer.

src/Foundatio.Logging.Xunit/TestLoggerFactory.cs
StephenCleary: Is there any reason we *wouldn't* want trace information by default?

niemyjski: I'd say the biggest thing is it can slow down your tests big time as everything gets that min log level. If we set the min log level to info. Then in the specific Test we can do Log.SetMinLevel<Queue>(Trace) and then we won't get trace info for caches etc.. This also makes looking through the build log much smaller.

src/Foundatio/Utility/ScheduledTimer.cs
niemyjski: Also thinking more on this, could be a ton of allocations / work creating a new timer all the time.

src/Foundatio.Logging.Xunit/TestLoggerFactory.cs
StephenCleary: Fair enough; I'll change it back.

src/Foundatio/Metrics/MetricTimer.cs
StephenCleary: This kind of stopwatch doesn't "run" - it just captures the current time.

niemyjski: That's not guaranteed, unless the default package doesn't use a stopwatch (https://github.com/Reactive-Extensions/Rx.NET/blob/1ad21f5a853f2b7be5f3a7e4a8d276e7d28f0d73/Rx.NET/Source/System.Reactive.Core/Reactive/Concurrency/Stopwatch.Default.cs)

src/Foundatio/Utility/ScheduledTimer.cs
StephenCleary: I don't think it'll cause any race conditions that aren't already there. To be honest, though, I think this entire class could use a rewrite.

src/Foundatio/Metrics/MetricTimer.cs
StephenCleary: There isn't a `Stop` or `Dispose`, so if it's a problem, it's Rx's problem.

niemyjski: Seems the interface doesn't have a stop. You would think you would want to stop that from running.

niemyjski: https://github.com/Reactive-Extensions/Rx.NET/issues/278

src/Foundatio/Utility/ScheduledTimer.cs
niemyjski: Feel free to take a stab at it :)

ejsmith:
Sorry I haven't had time to look at this yet. Let's get together and go over the changes soon.

src/Foundatio/Utility/ScheduledTimer.cs
```diff
@@ -25,8 +27,8 @@ public class ScheduledTimer : IDisposable {
             _timerCallback = timerCallback;
             _minimumInterval = minimumIntervalTime ?? TimeSpan.Zero;
 
-            int dueTimeMs = dueTime.HasValue ? (int)dueTime.Value.TotalMilliseconds : Timeout.Infinite;
-            _timer = new Timer(s => RunCallbackAsync().GetAwaiter().GetResult(), null, dueTimeMs, Timeout.Infinite);
+            if (dueTime != null)
+                _timer = SystemClock.Instance.Schedule(dueTime.Value, async () => await RunCallbackAsync());
```
ejsmith: It doesn't look like the Schedule method has any overloads that take tasks and the code seems to be calling the overload that just takes an Action. Wouldn't that mean that this would be an async void call?

src/Foundatio/Utility/SystemClock.cs
```diff
@@ -1,175 +1,118 @@
...
+            return new CancellationTokenSource(timeout);
         }
+    }
 
-        public void Sleep(int milliseconds) {
-            if (!_fakeSleep) {
-                Thread.Sleep(milliseconds);
-                return;
+    public static class SystemClock {
+        private static readonly AsyncLocal<ISystemClock> _instance = new AsyncLocal<ISystemClock>();
```
ejsmith: I'm not really sure what this is accomplishing. The class and instance are static. So it's not like the instance is scoped to only callers inside of the using blocks.

src/Foundatio/Utility/SystemClock.cs
```diff
@@ -1,175 +1,118 @@
...
-    public interface ISystemClock {
-        DateTime Now();
-        DateTime UtcNow();
-        DateTimeOffset OffsetNow();
-        DateTimeOffset OffsetUtcNow();
-        void Sleep(int milliseconds);
-        Task SleepAsync(int milliseconds, CancellationToken ct);
-        TimeSpan TimeZoneOffset();
+    public interface ISystemClock : IScheduler
+    {
```
ejsmith: Please reformat this to match the code styling for the rest of the lib.

niemyjski:
We've decided at least for the short term not to bring in the rx library. We've merged in most of the other changes from this pr in the dotnet branch. We greatly appreciate the pr

PR #63, opened 2016-09-21
Updated link to active Metrics.NET repo

slolife:
Readme on https://github.com/etishor/Metrics.NET says that it is no longer maintained, but that https://github.com/Recognos/Metrics.NET is actively maintained.


niemyjski:
Thank you!

PR #66, opened 2016-09-22
Adding ICacheClient.GetOrAddAsync

dcomartin:
Was looking at other In-Memory caching solutions and found a GetOrAdd was useful.  Ultimately I could of just made this into an extension method to use privately but figured I'd throw it in here if you thought it would be worth adding.


niemyjski:
My vote is to keep the interface clean and make this into an extension method. If you want to create an extension method for this and keep your unit test I'd be more than happy to accept this.

niemyjski:
I'm going to close this, would you please add an extension method for this instead.

PR #69, opened 2016-09-26
Cache all values in the local cache rather than just the serialized ones.

lynnroth:
Cache all values in the local cache rather than just the serialized ones.

Fix test to match counts.

### DIFF
--- a/src/Foundatio/Caching/HybridCacheClient.cs
+++ b/src/Foundatio/Caching/HybridCacheClient.cs
@@ -88,19 +88,17 @@ namespace Foundatio.Caching {
             CacheValue<T> cacheValue;
             bool requiresSerialization = TypeRequiresSerialization(typeof(T));
             _logger.Trace("Type requires serialization: {0}", requiresSerialization);
-
-            if (requiresSerialization) {
-                cacheValue = await _localCache.GetAsync<T>(key).AnyContext();
-                if (cacheValue.HasValue) {
-                    _logger.Trace("Local cache hit: {0}", key);
-                    Interlocked.Increment(ref _localCacheHits);
-                    return cacheValue;
-                }
+            
+            cacheValue = await _localCache.GetAsync<T>(key).AnyContext();
+            if (cacheValue.HasValue) {
+                _logger.Trace("Local cache hit: {0}", key);
+                Interlocked.Increment(ref _localCacheHits);
+                return cacheValue;
             }
 
             _logger.Trace("Local cache miss: {0}", key);
             cacheValue = await _distributedCache.GetAsync<T>(key).AnyContext();
-            if (requiresSerialization && cacheValue.HasValue) {
+            if (cacheValue.HasValue) {
                 var expiration = await _distributedCache.GetExpirationAsync(key).AnyContext();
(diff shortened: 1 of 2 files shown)


src/Foundatio/Caching/HybridCacheClient.cs
```diff
@@ -88,19 +88,17 @@ public class HybridCacheClient : ICacheClient {
             CacheValue<T> cacheValue;
             bool requiresSerialization = TypeRequiresSerialization(typeof(T));
```
niemyjski: We should be able to get rid of this as it's no longer being used? If this method isn't being used anymore can we remove it as well.

lynnroth: I would agree.  It's currently only used for the logging.  If that isn't needed, we can remove it.  I'll update the PR.

niemyjski:
LGTM CC @ejsmith 

I wish I remember why we did this but all tests pass..

ejsmith:
Yeah, I'm a little nervous about this change since we specifically added this behavior at one point. I think we need to do some more testing on this to be sure.

lynnroth:
What if it were an optional configuration where you had to turn on "CacheAllLocal" or "CacheSimpleObjectsLocal".  Simple boolean property and basically add back the stuff I removed and include that boolean in the check?

ejsmith:
I want to try and figure out why we excluded those before and if that reason is still relevant. It would be nice to do some benchmarks for this stuff to try and see if we are actually improving perf with the in memory stuff. It might have been us saying that your spending more time serializing and storing the value in 2 locations than it was to just store the simple value in the real cache.

lynnroth:
I did some performance checks on caching everything local vs not.  With a local redis (on windows) vs a remote redis (on Linux running on VMware over my gigabit link) both version 3.2.

Redis running locally runs in about 41% of the time as compared to redis running remotely.
With Redis running locally, caching all (with local cache size turned up to handle everything) runs in about 70% of the time as compared to not caching simple values locally.
With Redis running remotely,  the ratio is approximately the same.

So, running redis remotely and not caching simple values have about the same hit...

This was with caching ints.  Nothing more, so fairly simple test, but seems to indicate that things run quicker if the local cache can handle the size.

ejsmith:
That's awesome @lynnroth ! It would be really nice to put together a benchmark.net test for this with the code you made. Have you ever used that?

lynnroth:
I have not used benchmark.net.  seen it, but not used it.

niemyjski:
I'm curious what is the status of this pr?

lynnroth:
Any chance this can be merged at some point?

niemyjski:
Sorry, I wasn't sure about what the status was previously. I'll merge it in, we'll just have to keep an eye on any side effects (I'm sure we had this check there for a reason but we don't remember).

lynnroth:
Thanks!

niemyjski:
awe crap, we merged this into the wrong branch.

niemyjski:
@lynnroth  can you please resubmit this into the master branch. I'm going to be working on it real soon and we aren't 100% sure if we are going to merge in the rx branch

lynnroth:
New PR https://github.com/exceptionless/Foundatio/pull/77

PR #70, opened 2016-09-30
Extension method to ICachingClient for GetOrAddAsync

dcomartin:
Per recommendation, which I think made way more sense, I added this as an extension method.  Kept all the other tests that I had in place.

### DIFF
--- a/src/Foundatio/Caching/CacheClientHelpers.cs
+++ b/src/Foundatio/Caching/CacheClientHelpers.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Foundatio.Caching
+{
+    public static class CacheClientHelpers
+    {
+        public static async Task<CacheValue<T>> GetOrAddAsync<T>(this ICacheClient client, string key, Func<T> addFunc,
+            TimeSpan? expiresIn = null)
+        {
+            var cachedValue = await client.GetAsync<T>(key);
+            if (cachedValue.HasValue) return cachedValue;
+
+            var value = addFunc();
+
+            var addResult = await client.AddAsync(key, value, expiresIn);
+            return addResult ? new CacheValue<T>(value, true) : CacheValue<T>.NoValue;
+        }
+    }
+}
--- a/test/Foundatio.Redis.Tests/Caching/RedisCacheClientTests.cs
+++ b/test/Foundatio.Redis.Tests/Caching/RedisCacheClientTests.cs
@@ -72,6 +72,12 @@ namespace Foundatio.Redis.Tests.Caching {
             await base.CanManageSets();
         }
 
+        [Fact]
+        public override Task CanGetOrAddAsync()
+        {
+            return base.CanGetOrAddAsync();
+        }
+
         [Fact(Skip = "Performance Test")]
         public override Task MeasureThroughput() {
             return base.MeasureThroughput();
@@ -86,5 +92,7 @@ namespace Foundatio.Redis.Tests.Caching {
         public override Task MeasureSerializerComplexThroughput() {
             return base.MeasureSerializerComplexThroughput();
         }
+
+        
     }
 }
--- a/test/Foundatio.Redis.Tests/Caching/RedisHybridCacheClientTests.cs
+++ b/test/Foundatio.Redis.Tests/Caching/RedisHybridCacheClientTests.cs
@@ -53,6 +53,12 @@ namespace Foundatio.Redis.Tests.Caching {
         }
 
         [Fact]
+        public override Task CanGetOrAddAsync()
+        {
+            return base.CanGetOrAddAsync();
+        }
+
+        [Fact]
         public override Task WillUseLocalCache() {
             return base.WillUseLocalCache();
         }
@@ -81,5 +87,6 @@ namespace Foundatio.Redis.Tests.Caching {
         public override Task MeasureSerializerComplexThroughput() {
             return base.MeasureSerializerComplexThroughput();
         }
+
     }
 }
--- a/test/Foundatio.Tests/Caching/CacheClientTestsBase.cs
+++ b/test/Foundatio.Tests/Caching/CacheClientTestsBase.cs
@@ -23,6 +23,30 @@ namespace Foundatio.Tests.Caching {
             return null;
         }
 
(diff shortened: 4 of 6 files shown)


src/Foundatio/Caching/CacheClientHelpers.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Foundatio.Caching
```
niemyjski: I think we already have a set of extension methods in the same file as the interface

src/Foundatio/Caching/CacheClientHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public static class CacheClientHelpers
+    {
+        public static async Task<CacheValue<T>> GetOrAddAsync<T>(this ICacheClient client, string key, Func<T> addFunc,
+            TimeSpan? expiresIn = null)
```
niemyjski: I'm on my mobile device and curious what the concurrent collections getoradd signature looks like? Can we also add some argument validation

src/Foundatio/Caching/CacheClientHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+        {
+            var cachedValue = await client.GetAsync<T>(key);
+            if (cachedValue.HasValue) return cachedValue;
+
+            var value = addFunc();
+
```
niemyjski: What if the func is null or throws an exception? Should we let it bubble up or add a null cache key?

src/Foundatio/Caching/CacheClientHelpers.cs
```diff
@@ -0,0 +1,20 @@
...
+            var addResult = await client.AddAsync(key, value, expiresIn);
+            return addResult ? new CacheValue<T>(value, true) : CacheValue<T>.NoValue;
```
niemyjski: If the cache value already exists (race) between get and add this could return false.  Maybe we should call set and check this results?? Thoughts?

test/Foundatio.Redis.Tests/Caching/RedisCacheClientTests.cs
```diff
@@ -72,6 +72,12 @@ public class RedisCacheClientTests : CacheClientTestsBase {
             await base.CanManageSets();
         }
 
+        [Fact]
+        public override Task CanGetOrAddAsync()
+        {
```
niemyjski: Would it be possible to match our bracket style

src/Foundatio/Caching/CacheClientHelpers.cs
niemyjski: Or returns null

test/Foundatio.Redis.Tests/Caching/RedisCacheClientTests.cs
dcomartin: Ya, my bad, never even noticed.

src/Foundatio/Caching/CacheClientHelpers.cs
dcomartin: ICacheClient.cs just defines the interface.  I can put the class/extension in that file if you prefer.

src/Foundatio/Caching/CacheClientHelpers.cs
dcomartin: I originally had arg check but then removed it because `GetAsync<T>` does a `String.IsNullOrEmpty(key)`.  I can do the same check prior if you.

I'll take a look at the concurrent collections one, let me get back to that.  I originally got this idea for the GetOrAdd with the `Func<T>` because of looking at another library LazyCache.

src/Foundatio/Caching/CacheClientHelpers.cs
dcomartin: I think it returning null is ok, however the func param being null needs to throw I think.

src/Foundatio/Caching/CacheClientHelpers.cs
dcomartin: Never noticed or knew what the `Set` did?  Upsert?

src/Foundatio/Caching/CacheClientHelpers.cs
dcomartin: `ConcurrentDictionary<TKey, TValue>.GetOrAdd Method (TKey, TValue)`

https://msdn.microsoft.com/en-us/library/ee378674(v=vs.110).aspx

src/Foundatio/Caching/CacheClientHelpers.cs
niemyjski: Yeah, I think it would be good to have it right next to the interface (it's what we've been doing with all the other abstractions)

https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio/Queues/IQueue.cs#L48

src/Foundatio/Caching/CacheClientHelpers.cs
niemyjski: Yeah, its best if it doesn't have to create another async state machine behind the scenes just to throw an error :), unless anyone disagrees :)

https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio/Caching/InMemoryCacheClient.cs#L99

src/Foundatio/Caching/CacheClientHelpers.cs
niemyjski: I think if add returns null, maybe we don't call set, we just return.

src/Foundatio/Caching/CacheClientHelpers.cs
niemyjski: Yeah, it upserts :)

dcomartin:
I'll close this and until I can get around to looking at this again.

niemyjski:
@dcomartin My apologies, if you want to fix the bracing styles and what we talked about, we'd be more than happy to accept this.

niemyjski:
Do you want me to reopen it, or is there anything I can do to help?

niemyjski:
@dcomartin can you please resubmit this into the master branch. I'm going to be working on it real soon and we aren't 100% sure if we are going to merge in the rx branch

PR #84, opened 2017-03-02
Feature reduced locking

niemyjski:


src/Foundatio/Queues/QueueBase.cs
```diff
@@ -82,31 +86,43 @@ public virtual Task<IQueueEntry<T>> DequeueAsync(TimeSpan? timeout = null)
         public AsyncEvent<EnqueuedEventArgs<T>> Enqueued { get; } = new AsyncEvent<EnqueuedEventArgs<T>>(true);
 
         protected virtual Task OnEnqueuedAsync(IQueueEntry<T> entry) {
+            if (Enqueued == null)
```
jamie94bc: The intention of these additional null checks is to save the overhead of creating the `EnqueuedEventArgs`? 

When written like it is currently, both the additional check and the null propagation operator are required for thread safety, but it could be rewritten as below?
```csharp
var enqueued = Enqueued;
if (enqueued == null) {
    return Task.CompletedTask;
}

var args = new EnqueuedEventArgs<T> { Queue = this, Entry = entry };
return enqueued.InvokeAsync(this, args);
```

niemyjski: Yeah nice catch I'll make this change

src/Foundatio.Redis/Messaging/RedisMessageBus.cs
```diff
@@ -30,7 +30,7 @@ public class RedisMessageBus : MessageBusBase, IMessageBus {
                     return;
 
                 _logger.Trace("Subscribing to topic: {0}", _topic);
-                _subscriber.Subscribe(_topic, OnMessage);
+                _subscriber.Subscribe(_topic, OnMessage, CommandFlags.FireAndForget);
```
ejsmith: This seems pretty iffy. I would want the messagebus to blow up if I did subscribe and it failed to subscribe.

PR #85, opened 2017-03-04
add amazon sqs support

pwelter34:


src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+using Amazon.Runtime;
+using Amazon.SQS;
+using Amazon.SQS.Model;
+using Foundatio.Logging;
+using Foundatio.Queues;
+using Foundatio.Serializer;
+using Foundatio.Utility;
+using Nito.AsyncEx;
+
+namespace Foundatio.AWS.Queues {
```
niemyjski: All our other queues are in the `Foundatio.Queues` namespace

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+            _readQueueTimeout = readQueueTimeout ?? TimeSpan.FromSeconds(2);
+
+            _client = new Lazy<AmazonSQSClient>(() => new AmazonSQSClient(_credentials, _regionEndpoint));
+        }
+
+        protected override async Task EnsureQueueCreatedAsync(CancellationToken cancellationToken = new CancellationToken()) {
+            if (!String.IsNullOrEmpty(_queueUrl))
+                return;
+
+            using (await _lock.LockAsync(cancellationToken).ConfigureAwait(false)) {
```
niemyjski: We have an AnyContext() extension method for ConfigureAwait(false)

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+            if (!await OnEnqueuingAsync(data).ConfigureAwait(false))
+                return null;
+
+            Interlocked.Increment(ref _enqueuedCount);
+            var message = new SendMessageRequest {
+                QueueUrl = _queueUrl,
+                MessageBody = await _serializer.SerializeToStringAsync(data).ConfigureAwait(false),
+            };
+
+            var response = await _client.Value.SendMessageAsync(message).ConfigureAwait(false);
```
niemyjski: Should we be retrying this? We retry in half of our current implementations (curious what your thoughts are) https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio.Redis/Queues/RedisQueue.cs#L201

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+
+                _queueUrl = createResponse.QueueUrl;
+            }
+        }
+
+        protected override async Task<string> EnqueueImplAsync(T data) {
+            if (!await OnEnqueuingAsync(data).ConfigureAwait(false))
+                return null;
+
+            Interlocked.Increment(ref _enqueuedCount);
```
niemyjski: In our other clients we do this after we publish the message.

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+            var response = await _client.Value.SendMessageAsync(message).ConfigureAwait(false);
+
+            var entry = new QueueEntry<T>(response.MessageId, data, this, SystemClock.UtcNow, 0);
+            await OnEnqueuedAsync(entry).ConfigureAwait(false);
+
+            return response.MessageId;
+        }
+
+        protected override async Task<IQueueEntry<T>> DequeueImplAsync(CancellationToken cancellationToken) {
+            var visibilityTimeout = Convert.ToInt32(_workItemTimeout.TotalSeconds);
```
niemyjski: It might be better to do conversions in the constructor instead of every time we dequeue it

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+
+
+            // sqs doesn't support already canceled token, change timeout and token for sqs pattern
+            var waitTimeout = cancellationToken.IsCancellationRequested
+                ? 0
+                : Convert.ToInt32(_readQueueTimeout.TotalSeconds);
+
+            var cancel = cancellationToken.IsCancellationRequested
```
niemyjski: In our other queue implementations (azure, redis and others) we use a linked cancellation token and async auto reset events (https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio.AzureStorage/Queues/AzureStorageQueue.cs#L24) which we cancel in dispose.. We also use a do-while loop in at least 3 of our clients (pretty sure all off them do) 

https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio.AzureStorage/Queues/AzureStorageQueue.cs#L93

https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio.Redis/Queues/RedisQueue.cs#L264

You can run into some edge cases where you need to wait for a message to be received instead of returning immediately.  I guess if all tests pass it's good,

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+                WaitTimeSeconds = waitTimeout,
+                AttributeNames = new List<string> { "ApproximateReceiveCount", "SentTimestamp" }
+            };
+
+            ReceiveMessageResponse response = null;
+
+            try {
+                response = await _client.Value.ReceiveMessageAsync(request, cancel).ConfigureAwait(false);
+            }
+            catch (TaskCanceledException) {
```
niemyjski: @StephenCleary recommended we catch the base `OperationCanceledException` exception for all cancellations.

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+            _enqueuedCount = 0;
+            _dequeuedCount = 0;
+            _completedCount = 0;
+            _abandonedCount = 0;
+            _workerErrorCount = 0;
+        }
+
+        protected override void StartWorkingImpl(Func<IQueueEntry<T>, CancellationToken, Task> handler, bool autoComplete, CancellationToken cancellationToken) {
+            Task.Run(async () => {
+                while (!cancellationToken.IsCancellationRequested) {
```
niemyjski: we use the linked cancellation token here too, otherwise this task might never end (especially in unit tests) when the test is disposed

https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio.Redis/Queues/RedisQueue.cs#L217

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,269 @@
...
+                    IQueueEntry<T> entry = null;
+                    try {
+                        entry = await DequeueImplAsync(cancellationToken).ConfigureAwait(false);
```
niemyjski: We seem to be breaking this apart into multiple try catches and checking the token to save work

https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio.Redis/Queues/RedisQueue.cs#L226
https://github.com/exceptionless/Foundatio/blob/master/src/Foundatio.AzureStorage/Queues/AzureStorageQueue.cs#L185

Is there any chance you could update the redis implementation (first try catch block should be catching OperationCanceledException)?

test/Foundatio.AWS.Tests/Foundatio.AWS.Tests.csproj
```diff
@@ -4,10 +4,15 @@
     <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\src\Foundatio\Foundatio.csproj" />
     <ProjectReference Include="..\..\src\Foundatio.AWS\Foundatio.AWS.csproj" />
+    <ProjectReference Include="..\..\src\Foundatio.Logging.Xunit\Foundatio.Logging.Xunit.csproj" />
     <ProjectReference Include="..\Foundatio.Tests\Foundatio.Tests.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="log4net" Version="2.0.8" />
```
niemyjski: These references are brought into automatically via our tests.props

test/Foundatio.Tests/Queue/QueueTestBase.cs
```diff
@@ -28,6 +28,10 @@ public abstract class QueueTestBase : TestWithLoggingBase, IDisposable {
             return null;
         }
 
+        protected virtual Task CloseQueue(IQueue<SimpleWorkItem> queue) {
```
niemyjski: I wonder if our other azure queues and rabbit queues are being cleaned up, probably not :\. I'm trying to think in the tests where we create multiple queue instances if they would need to be cleaned up or not. I almost think they wouldn't as they should all have the same queue name but different instance ids.

src/Foundatio.AWS/Queues/SQSQueue.cs
niemyjski: Looks like this one was still missed :)

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        protected override async Task<IQueueEntry<T>> DequeueImplAsync(CancellationToken cancellationToken) {
+
+            var visibilityTimeout = _workItemTimeoutSeconds;
+
+
+            // sqs doesn't support already canceled token, change timeout and token for sqs pattern
+            var waitTimeout = cancellationToken.IsCancellationRequested ? 0 : _readQueueTimeoutSeconds;
+
+            var cancel = cancellationToken.IsCancellationRequested
```
niemyjski: I don't see this being used anywhere

src/Foundatio.AWS/Queues/SQSQueue.cs
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        public override void Dispose() {
+            _logger.Trace("Queue {0} dispose", _queueName);
+
+            if (_client.IsValueCreated)
+                _client.Value.Dispose();
+
+            base.Dispose();
+        }
```
niemyjski: The disposed cancellation token isn't being cancelled here.

test/Foundatio.AWS.Tests/Foundatio.AWS.Tests.csproj
niemyjski: I don't think you meant to bring in log4net

src/Foundatio.AWS/Queues/SQSQueue.cs
pwelter34: That one is not supported, it returns something not supported by the extension method

test/Foundatio.AWS.Tests/Foundatio.AWS.Tests.csproj
pwelter34: aws sdk looks for it via reflection.  i get false errors if its missing.

src/Foundatio.AWS/Queues/SQSQueue.cs
pwelter34: fixed

src/Foundatio.AWS/Queues/SQSQueue.cs
pwelter34: fixed

src/Foundatio.AWS/Queues/SQSQueue.cs
niemyjski: We should have an extension method for that, maybe it's missing from the project?

niemyjski:
Looks good to me!!! Thanks for the contrib

ejsmith:
I think the tests fail on this because it won't give it access to the secrets for SQS access.

PR #89, opened 2017-03-16
CloudWatch metrics

ejsmith:


jamie94bc:
@niemyjski I'm okay with making `IQueueJob` generic, makes sense.

Issue #90, opened 2017-03-23
ASB MessageBus completes a poisonous message rather than DLQ it.


SeanFeldman:
When `AzureServiceBusMessageBus` receives a poisonous message, it [completes it](https://github.com/exceptionless/Foundatio/blob/d2967a1f56d8d3fa31cc194bc260561f153e2ac1/src/Foundatio.AzureServiceBus/Messaging/AzureServiceBusMessageBus.cs#L44) with a log entry. ASB supports deadletter queues. Poisonous messages should be forwarded to the DLQ rather being completed.

SeanFeldman:
I can submit a PR, but not sure if there's a test case that already covers this or not.

niemyjski:
@SeanFeldman Would it be better to just move the current message to deadletter queue on error instead of submitting the log message?

niemyjski:
I also have a massive amount of changes in the cloud watch branch (where we've been standardizing mb and queue creation.

SeanFeldman:
@niemyjski log entry was in the original code. It makes sense to write a log to allow monitoring to detect that things went wrong. Perhaps lowering threshold from Error to a Warning.

niemyjski:
Thanks, I'll take a look into this :) Is there any chance you could update your pr to the cloud watch branch as I'm going to have merge conflicts as is :)

SeanFeldman:
Done

niemyjski:
Thanks! I'll play around with this, it would be good to have serialization tests for all our on message implementations. Guess we'll add that to the todo :)

PR #91, opened 2017-03-23
Asb deadletter poisonous messages

SeanFeldman:
Addresses issue #90 

Poisonous message is deadlettered instead of just being completed.
No tests provided as it would require to spoof a poisonous `BrokeredMessage` rather than following what the tests are doing now (using real messages delivered by the broker).


SeanFeldman:
Closed in favor of #92

PR #100, opened 2017-07-20
adding delete files by search

ttugrad01:


src/Foundatio/Storage/ActionableStream.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Foundatio.Storage {
+    public class ActionableStream : Stream {
+        private readonly Action _disposeAction;
+        private readonly Stream _stream;
+
+        protected override void Dispose(bool disposing) {
+            try {
+                _disposeAction.Invoke();
+            } catch { /* ignore if these are already disposed;  this is to make sure they are */ }
```
niemyjski: You could add a flag here` _isDisposed` and git rid of this try catch.

src/Foundatio/Storage/ActionableStream.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.IO;
+
+namespace Foundatio.Storage {
+    public class ActionableStream : Stream {
```
niemyjski: I don't see this being used anywhere.

PR #101, opened 2017-07-25
Signed Foundatio Packages

rerikson:
This commit adds signed packages for all of the projects in the Foundatio repo. Unfortunately some packages that Foundatio reference do not provide signed packages, like Nito.AsyncEx.Coordination.  This is not a problem for consumers using CoreCLR, but for those using the DesktopCLR it will cause a runtime issue as discussed [here](https://stackoverflow.com/questions/33760037/why-does-roslyn-give-warning-cs8002-instead-of-error-cs1577-for-unsigned-referen)

Users of the DesktopCLR who want to reference the signed packages would need to use a tool like [strongnamer](https://github.com/dsplaisted/strongnamer) which provides some build targets that automatically sign all packages.


niemyjski:
Should public signing only happen on NT? The changes look good, please let me know when it's ready :)

rerikson:
It's actually only not happening on NT.  I'm not sure if we need that or not, I borrowed it from the Exceptionless signing process. 

I would say the code is ready now, but after investigating why my previous commit caused the [build to fail ](https://ci.appveyor.com/project/Exceptionless/foundatio/build/5.1.1461-gergibre)I think we should reconsider having separate projects that end in ".Signed".  You're correct when you said that a strong-name package can reference a non-strong-name package [here](https://github.com/FoundatioFx/Foundatio.Redis/pull/2).  But, that is only correct when running an app  against CoreCLR.

[MSDN](https://docs.microsoft.com/en-us/dotnet/framework/app-domains/create-and-use-strong-named-assemblies) still says that strong-name assemblies can only use types from non-strong-name assemblies, but as you and [this roslyn issue](https://github.com/dotnet/roslyn/issues/7277) point out, that's only true now for apps running against the DesktopCLR.

So, the reason I recommend having the separate ".Signed" packages is because users running the full DesktopCLR will experience a runtime error if we just sign the original packages.  To avoid the runtime error everyone running the DesktopCLR would need to use a tool like the strongnamer tool I mentioned in the original PR.

niemyjski:
I think we may need to source import or get @StephenCleary to strong name his assemblies :). I can merge this into a branch if you'd like until upstream dependencies get updated

rerikson:
Merging it into a branch sounds good. Thanks :)

niemyjski:
Awesome, It all looks good. I think the only thing left would be to copy the license file from the nito libs and just put them into the folder (so they are visible). Also, I'm thinking probably would be nice to remove any nito classes that aren't being used but we can always do that later!

src/Foundatio.Jobs.Commands/Foundatio.Jobs.Commands.csproj
```diff
@@ -4,6 +4,11 @@
     <TargetFrameworks>net46</TargetFrameworks>
     <PackageTags>Distributed;Jobs;NETSTANDARD;Core</PackageTags>
   </PropertyGroup>
+  <PropertyGroup Label="Signing">
+    <SignAssembly>true</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+    <AssemblyOriginatorKeyFile>..\Exceptionless.snk</AssemblyOriginatorKeyFile>
```
niemyjski: We need to copy the strong name file over, I'm currently not sure how the build is passing (it should fail if the strong named file doesn't exist).

appveyor.yml
```diff
@@ -14,16 +14,19 @@ init:
...
 before_build:
   - ps: .\build\Set-BuildVersion -Version $env:APPVEYOR_BUILD_VERSION -Suffix $env:VERSION_SUFFIX
-  - appveyor-retry dotnet restore -v Minimal
+  - appveyor-retry dotnet restore -v Minimal  
 
 build_script:
-  - ps: dotnet build -c Release
+  - ps: dotnet build -c Release 
 
+test_script:
```
niemyjski: Did you come across an issue with unit tests? They should be picked up automatically via appveyor. The test harness is a shared library that the other foundatio libs build on, I believe (@ejsmith can confirm) that only the Foundation.Tests are the only runable tests?

src/Foundatio.Jobs.Commands/Foundatio.Jobs.Commands.csproj
niemyjski: I'm thinking this wasn't copied or added because of a git ignore.. You might have to do a `git add PATHTOSNK -f`

appveyor.yml
rerikson: Yeah, I did run into an issue when I was trying to run them with dotnet test.  It worked inside of Visual Studio, but I wanted to make sure it would work if appveyor was using dotnet test internally.  The error I got was "Make sure test project has a nuget reference of package "microsoft.testplatform.testhost".  I'll go ahead and revert it to the original version and see how that goes.

src/Foundatio.Jobs.Commands/Foundatio.Jobs.Commands.csproj
rerikson: Are you sure it's not being copied over?  The reason I ask is that it is included in the [fork I've been working off of](https://github.com/TicketVista/Foundatio/tree/master/src) and it looks like it's included in the PR in the file above this file we're commenting on.

niemyjski: weird, I wasn't seeing it in the pr files before :\

test/Foundatio.TestHarness/Foundatio.TestHarness.csproj
```diff
@@ -1,14 +1,24 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.5;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard1.5;netcoreapp1.1;net46</TargetFrameworks>
```
niemyjski: The pr looks good, we might have to revert the netcoreapp1.1 later but it's all good.

niemyjski:
LGTM, Thanks for the PR! I'll get this merged in today unless anyone speaks up :)

test/Foundatio.TestHarness/Foundatio.TestHarness.csproj
ejsmith: Shouldn't have to target both netstandard and netcoreapp they would be testing against the same platform.

PR #102, opened 2017-07-30
Feature/storage delete files

ttugrad01:


src/Foundatio/Storage/ActionableStream.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        public override long Length => _stream.Length;
+
+        public override long Position {
+            get => _stream.Position;
+            set => _stream.Position = value;
+        }
+
+        public override int Read(byte[] buffer, int offset, int count) {
```
niemyjski: Are there also async overloads for stream?

src/Foundatio/Storage/IFileStorage.cs
```diff
@@ -84,5 +85,43 @@ public static class FileStorageExtensions {
         public static Task<bool> SaveFileAsync(this IFileStorage storage, string path, string contents) {
             return storage.SaveFileAsync(path, new MemoryStream(Encoding.UTF8.GetBytes(contents ?? String.Empty)));
         }
+
+        internal static bool ContainsWildcard(this string searchPattern) {
```
niemyjski: I could have sworn we had a string extension for this in foundatio, but maybe that's just in the exceptionless project.

niemyjski:
Left some comments about async overloads but looks good to me

PR #103, opened 2017-07-31
removing IsFolderSearch() extension

ttugrad01:


src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -153,7 +153,7 @@ public class FolderFileStorage : IFileStorage {
                 return Task.CompletedTask;
             }
             
-            if (searchPattern.IsFolderSearch()) {
+            if (!searchPattern.IsFileSearch()) {
```
ejsmith: Still need to normalize slashes. Looks like it was a bug from before that we weren't doing that.

I think this case should be checking for ends with / or /* or Directory.Exists for the full path string is true.

PR #105, opened 2017-08-09
Synchronize RenewedTimeUtc when dequeued from InMemoryQueue

edwardmeng:
Synchronize RenewedTimeUtc when dequeued from InMemoryQueue, to avoid abandoned incorrectly in the maintain job. This bug will lead an error in exceptionless EventPostJob: When there are high throughput event posts, the job will throw exception while maintain job processing. The error message is: 
```
EventPostsJob Error processing EventPost queue entry (890ac28901f042f1beea9e5b2a8b5168). System.Exception: Unable to remove item from the dequeued list.
```
At the same time the maintain job writes the following log message:
```
INFO DoMaintenance Abandon: 890ac28901f042f1beea9e5b2a8b5168 
```

### DIFF
--- a/src/Foundatio/Queues/InMemoryQueue.cs
+++ b/src/Foundatio/Queues/InMemoryQueue.cs
@@ -151,6 +151,7 @@ namespace Foundatio.Queues {
             _logger.Trace("Dequeue: Got Item");
 
             var entry = new QueueEntry<T>(info.Id, info.Value.DeepClone(), this, info.EnqueuedTimeUtc, info.Attempts);
+            await entry.RenewLockAsync();
             await OnDequeuedAsync(entry).AnyContext();
             ScheduleNextMaintenance(SystemClock.UtcNow.Add(_options.WorkItemTimeout));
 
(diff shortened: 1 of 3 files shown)


niemyjski:
@edwardmeng would it be possible to create a test that reproduces this (if it's hard to reproduce then I understand). I would love to see if this effects any of our other impls.

Also, I'm trying to understand the full scope of this.. Would this only occur if you dequeue the item and the item being worked on takes a while to run and then maintenance runs?

edwardmeng:
I have created the unit test method for this scenario. 
There are more than two queue work items. Every time work item has been dequeued from the InMemoryQueue, the maintain job will schedule the next execution time. So when the first work item has been dequeued, and the last work item dequeued before the first maintain execution time, but try to complete after maintain job executed. The InMemoryQueue will have been abandoned the last work item due to incorrect RenewedTimeUtc value, then there will be an error thrown in the last work item complete process. The error message will be:
```
Unable to remove item from the dequeued list.
```
To fix this issue, just renew the RenewedTimeUtc for every work item when it dequeued from InMemoryQueue.

niemyjski:
Awesome! Thanks for the explanation, pr and tests! It's greatly appreciated!!

PR #106, opened 2017-08-14
netstandard1.5 for Foundatio.Logging.Serilog

pje-cloudbuy:
Support netstandard1.5 for Foundatio.Logging.Serilog.

In keeping with the rest of Foundatio, which supports both .NET Framework 4.6 and .NET Standard 1.5, add support for .NET Standard 1.5 to Foundatio.Logging.Serilog.

Simply adding this to the list of target frameworks appears to work and allow this to build to .NET Standard 1.5 locally in Visual Studio 2017.  Perhaps it was just an oversight that this has not been done already?

All logging unit-tests still pass.


niemyjski:
Thanks for the PR!

PR #107, opened 2017-09-14
convert to Microsoft.Extensions.Logging

pwelter34:


src/Foundatio/Jobs/IJob.cs
```diff
@@ -23,7 +24,7 @@ public static class JobExtensions {
             string jobName = job.GetType().Name;
             var logger = job.GetLogger();
 
-            using (logger.BeginScope(s => s.Property("job", jobName))) {
+            using (logger.BeginScope($"job: {jobName}")) {
```
pwelter34: Not sure what this should be

src/Foundatio/Jobs/JobRunner.cs
```diff
@@ -97,7 +98,7 @@ public JobRunner(Func<IJob> jobFactory, ILoggerFactory loggerFactory = null, Tim
             }
 
             _jobName = TypeHelper.GetTypeDisplayName(job.GetType());
-            using (_logger.BeginScope(s => s.Property("job", _jobName))) {
+            using (_logger.BeginScope($"job: {_jobName}")) {
```
pwelter34: again, not sure here

src/Foundatio/Logging/LoggerExtensions.cs
```diff
@@ -1,16 +0,0 @@
...
-using System.Collections.Generic;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
-
-namespace Foundatio.Logging {
-    public static class LoggerExtensions {
-        public static ILogger GetLogger(this object target) {
-            return target is IHaveLogger accessor ? accessor.Logger : NullLogger.Instance;
-        }
-
```
pwelter34: why remove BeginScope extension?  helps add a dictionary.

ejsmith: Because we are removing logging. Don’t think we should be adding random extensions to libs. Should stay focused on the services we are trying to provide. I actually created a bunch more extension methods for logging. Maybe we should create a separate lib.

onuralp:
Hi,

While I am reviewing the change I saw you guys checking LogLevel IsEnabled before logging. I think It is not necessary. Because Microsoft is already checking.

https://github.com/aspnet/Logging/blob/000418aca3936a317309352c0a627c1b512f2c95/src/Microsoft.Extensions.Logging/Logger.cs#L52

https://github.com/aspnet/Logging/blob/000418aca3936a317309352c0a627c1b512f2c95/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L104

ejsmith:
Right, but if the logging code is doing any sort of calculation or string allocation, especially in performance critical sections, then it’s better to do the check ourselves.

onuralp:
@ejsmith, In that case, you absolutely right. 
For example 
https://github.com/FoundatioFx/Foundatio/blob/logging/src/Foundatio/Caching/HybridCacheClient.cs#L53

But somewhere like this. It is not necessary https://github.com/FoundatioFx/Foundatio/blob/logging/src/Foundatio/Caching/HybridCacheClient.cs#L56

So maybe we can create an extension method for this.

src/Foundatio/Metrics/MetricTimer.cs
```diff
@@ -15,13 +16,13 @@ public class MetricTimer : IAsyncDisposable {
...
-        public Task DisposeAsync() {
+        public void Dispose() {
             if (_disposed)
-                return Task.CompletedTask;
+                return;
 
             _disposed = true;
             _stopWatch.Stop();
-            return _client.TimerAsync(_name, (int)_stopWatch.ElapsedMilliseconds);
+            _client.Timer(_name, (int)_stopWatch.ElapsedMilliseconds);
```
niemyjski: Should we call dispose on the client here as well?

niemyjski:
@onuralp I agree, I also wonder if the check is still valid because I know there is a lot of work going on with argument validation and the like that we skip completely by doing a quick dictionary lookup.

ejsmith:
Agreed @onuralp the check is not necessary there.

PR #111, opened 2017-11-27
Fix CanRunWorkItemWithMetricsAsync test and skip CanSendMultithreaded…

onuralp:
It includes a fix for the broken unit test. And skip CanSendMultithreaded test on StatsDMetricsTests. I add skip this because I couldn't find the issue. You may explain what it is for.


niemyjski:
Thanks for the PR What flakey test were you finding for stats d?

onuralp:
In that Test case, SystemClockTests doesn't use. And It took a long time to run. 
But my point is that test doesn't pass. It expects 100 message count but methods return 99 or 97.

niemyjski:
Hmm, Thanks for the heads up. I'll try to run it repeatedly locally.

PR #113, opened 2017-12-21
Added a task queue implementation.

niemyjski:
This can prevent running async code in sync places as well as offload work and prevent issues like: https://github.com/StackExchange/StackExchange.Redis/issues/759


niemyjski:
@ejsmith we probably want to ensure the TaskQueue is lazy initialized so any publishers don't have to take the hit of launching extra threads.

PR #116, opened 2018-01-15
Improve performance of FolderFileStorage and add loggings.

edwardmeng:


src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -7,15 +7,19 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Foundatio.Storage {
     public class FolderFileStorage : IFileStorage {
         private readonly object _lockObject = new object();
         private readonly ISerializer _serializer;
+        protected readonly ILogger _logger;
 
-        public FolderFileStorage(string folder, ISerializer serializer = null) {
+        public FolderFileStorage(string folder, ISerializer serializer = null, ILoggerFactory loggerFactory = null) {
```
niemyjski: We should create a Options class like the Queues, Cache and Message buses have to be consistent here.

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -39,7 +43,7 @@ public class FolderFileStorage : IFileStorage {
 
             try {
                 return Task.FromResult<Stream>(File.OpenRead(Path.Combine(Folder, path)));
-            } catch (FileNotFoundException) {
+            } catch (IOException ex) when(ex is FileNotFoundException || ex is DirectoryNotFoundException) {
                 return Task.FromResult<Stream>(null);
```
niemyjski: We should log a trace message here just so we can see more information if needed

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -77,22 +81,32 @@ public class FolderFileStorage : IFileStorage {
...
-                using (var fileStream = File.Create(file)) {
+                using (var fileStream = CreateFileStream(file)) {
                     if (stream.CanSeek)
                         stream.Seek(0, SeekOrigin.Begin);
 
                     stream.CopyTo(fileStream);
                 }
-            } catch (Exception) {
+            } catch (Exception ex) {
+                _logger.LogError(ex, "Error trying to save file '{path}'.", path);
```
niemyjski: Log messages are automatically quoted in structured logging:

`_logger.LogError(ex, "Error trying to save file: {Path}", path);`

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -77,22 +81,32 @@ public class FolderFileStorage : IFileStorage {
...
 
             string file = Path.Combine(Folder, path);
-            string directory = Path.GetDirectoryName(file);
-            if (directory != null)
-                Directory.CreateDirectory(directory);
 
             try {
-                using (var fileStream = File.Create(file)) {
+                using (var fileStream = CreateFileStream(file)) {
                     if (stream.CanSeek)
```
niemyjski: I can't see what method this is but I don't think we should be seeking any stream. We also probably need a null check on FileStream

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -77,22 +81,32 @@ public class FolderFileStorage : IFileStorage {
...
                         stream.Seek(0, SeekOrigin.Begin);
 
                     stream.CopyTo(fileStream);
```
niemyjski: Should this be a stream.CopyToAsync? Would this improve concurrent writes?

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -77,22 +81,32 @@ public class FolderFileStorage : IFileStorage {
...
                 return Task.FromResult(false);
             }
 
             return Task.FromResult(true);
+
+            Stream CreateFileStream(string filePath) {
+                try {
+                    return File.Create(filePath);
+                }
+                catch (DirectoryNotFoundException) {
```
niemyjski: What happens if the file already exists? I'm thinking this code could be outside of the catch,

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -119,7 +133,8 @@ public class FolderFileStorage : IFileStorage {
                         File.Move(oldFullPath, newFullPath);
                     }
                 }
-            } catch (Exception) {
+            } catch (Exception ex) {
+                _logger.LogError(ex, "Error trying to rename file '{path}' to '{newpath}'.", path, newpath);
```
niemyjski: See the log comment above, Should be proper cased and without quotes :).

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -157,7 +173,8 @@ public class FolderFileStorage : IFileStorage {
             
             try {
                 File.Delete(Path.Combine(Folder, path));
-            } catch (Exception) {
+            } catch (Exception ex) {
```
niemyjski: We should Ignore FileNotfound and DirectoryNotFound exceptions here.

src/Foundatio/Storage/FolderFileStorage.cs
edwardmeng: I don't know why seeking the stream, but the original code did this. But what should it to do with null check, throw exception or just log message and do nothing.

src/Foundatio/Storage/FolderFileStorage.cs
edwardmeng: I have checked for the APIs of file operations, the only is to use File.Exist check, but it leads more I/O operations.

src/Foundatio/Storage/FolderFileStorage.cs
niemyjski: this should be fine because the last file create is outside of the try and it will blow up how it used to.. We should see what removing this seek does, we shouldn't ever seek a stream.

src/Foundatio/Storage/FolderFileStorageOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Foundatio.Serializer;
+using Microsoft.Extensions.Logging;
+
+namespace Foundatio.Storage
+{
+    public class FolderFileStorageOptions
```
niemyjski: We should create a base options. Seems the commonality is between the implementations is Serializer and LoggerFactory

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -70,29 +78,39 @@ public class FolderFileStorage : IFileStorage {
...
+                _logger.LogError(ex, "Error trying to save file: {Path}", path);
+                return false;
             }
 
-            return Task.FromResult(true);
+
+            Stream CreateFileStream(string filePath) {
+                try {
+                    return new FileStream(filePath, FileMode.Create);
```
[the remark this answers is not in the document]
edwardmeng: I think it's the same as File.Create

src/Foundatio/Storage/FolderFileStorage.cs
edwardmeng: It will override the existing files.

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -157,7 +177,8 @@ public class FolderFileStorage : IFileStorage {
             
             try {
                 File.Delete(Path.Combine(Folder, path));
-            } catch (Exception) {
+            } catch (Exception ex) when (ex is FileNotFoundException || ex is DirectoryNotFoundException) {
+                _logger.LogError(ex, "Error trying to delete file: {Path}.", path);
```
niemyjski: I think this should be debug level. This isn't really an error you are trying to delete a file that doesn't exist.

src/Foundatio/Storage/FolderFileStorage.cs
niemyjski: Let's remove this seek.

src/Foundatio/Storage/FileStorageOptionsBase.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Foundatio.Serializer;
+using Microsoft.Extensions.Logging;
+
+namespace Foundatio.Storage {
+    public abstract class FileStorageOptionsBase {
```
niemyjski: Can we also quickly make this change to in memory storage as well (even if we don't add any logging for now,

PR #117, opened 2018-01-15
Exception issue

surgicalcoder:
Some minor cleaning up.

### DIFF
--- a/src/Foundatio/Storage/FolderFileStorage.cs
+++ b/src/Foundatio/Storage/FolderFileStorage.cs
@@ -152,7 +152,7 @@ namespace Foundatio.Storage {
                     File.Copy(Path.Combine(Folder, path), Path.Combine(Folder, targetPath));
                 }
             } catch (Exception ex) {
-                _logger.LogError(ex, "Error trying to copy file {Path} to {TargetPath}.", path, targetpath);
+                _logger.LogError(ex, "Error trying to copy file {Path} to {TargetPath}.", path, targetPath);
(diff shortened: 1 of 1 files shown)


src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -88,29 +88,19 @@ public class FolderFileStorage : IFileStorage {
             path = path.NormalizePath();
 
             string file = Path.Combine(Folder, path);
+            string directory = Path.GetDirectoryName(file);
+
+            Directory.CreateDirectory(directory);
```
niemyjski: We don't want to call this unless there is an exception on File.Create, this has overhead and we don't need it in most cases

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -88,29 +88,22 @@ public class FolderFileStorage : IFileStorage {
             path = path.NormalizePath();
 
             string file = Path.Combine(Folder, path);
+            string directory = Path.GetDirectoryName(file);
 
             try {
-                using (var fileStream = CreateFileStream(file)) {
+                using (var fileStream = File.Create(file)) {
```
niemyjski: I think using the existing method, is the best (we could move it to a private non local method). This could have recursion issues.

niemyjski:
Thanks for the PR, I made some of the changes and merged the branch into master. There was a bug in the code in this branch that could trigger a stack overflow with the recursive failure,

PR #120, opened 2018-01-27
Add fluent extension method for options classes so that we can create instances conveniently.

edwardmeng:


ejsmith:
This looks pretty good @edwardmeng. Only thing I was thinking different was to use a separate OptionsBuilder<T> where all the extension methods live off of that so that they don't pollute the actual options object itself:

```csharp
public class OptionsBuilder<T> where T: class, new() {
    public T Target { get; } = new T();
}

public class MyImp {
    private readonly MyImpOptions _options;

    public MyImp(ConfigureOptions<MyImpOptions> builder)
        : this(builder(new OptionsBuilder<MyImpOptions>()).Target) {}

    public MyImp(MyImpOptions options) {
        _options = options;
    }
}

public delegate OptionsBuilder<T> ConfigureOptions<T>(OptionsBuilder<T> builder) where T: class, new();

public class MyImpOptions {
    public string SomeOption { get; set; }
}

public static class MyImpOptionsBuilder {
    public static OptionsBuilder<MyImpOptions> WithSomeOption(this OptionsBuilder<MyImpOptions> builder, string someOption) {
        builder.Target.SomeOption = someOption;
        return builder;
    }
}

public static class Startup {
    public static void Main() {
        var myImp = new MyImp(b => b.WithSomeOption("value"));

        var myImp2 = new MyImp(new MyImpOptions { SomeOption = "value" });
    }
}
```

ejsmith:
I’ve always been torn on fluent builder methods being prefixed with something like `o => o.WithSomeOption()` vs just having `o => o.SomeOption()`.  What do you guys think @edwardmeng @niemyjski ?

edwardmeng:
@ejsmith I have completed the changes as your suggestion, please review again.

ejsmith:
Looks good except I like creating a delegate so that the ctors get a little simler and go from:

```csharp
public InMemoryFileStorage(Action<IOptionsBuilder<InMemoryFileStorageOptions>> config)
    : this(OptionsBuilder<InMemoryFileStorageOptions>.Build(config)) { }
```

to

```csharp
public delegate OptionsBuilder<T> ConfigureOptions<T>() where T: class, new();

public InMemoryFileStorage(ConfigureOptions<InMemoryFileStorageOptions> config)
    : this(OptionsBuilder<InMemoryFileStorageOptions>.Build(config)) { }
```

edwardmeng:
I have tried as your suggestion, but the ConfigureOptions cannot be defined covariant between base options and inherited options.

ejsmith:
Ok then I think it’s good to go.

PR #121, opened 2018-01-29
Trying a different approach for fluent builders

ejsmith:


src/Foundatio/Messaging/SharedMessageBusOptions.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using Foundatio.Serializer;
+using Foundatio.Utility;
+using Microsoft.Extensions.Logging;
+
+namespace Foundatio.Messaging {
+    public class SharedMessageBusOptions : SharedOptions {
```
edwardmeng: Shall we should keep the original class name `MessageBusOptionsBase` to avoid breaking changes for the implementation libraries.

src/Foundatio/Utility/OptionsBuilder.cs
```diff
@@ -1,17 +1,28 @@
 ﻿using System;
 
 namespace Foundatio {
-    public interface IOptionsBuilder<out T> {
-        T Target { get; }
+    public interface IOptionsBuilder {
+        object Target { get; }
+    }
+
+    public interface IOptionsBuilder<T> : IOptionsBuilder {
```
edwardmeng: The generic argument should be with `out` modifier to be covariant.

src/Foundatio/Utility/SharedOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+using Foundatio.Serializer;
+using Microsoft.Extensions.Logging;
+
+namespace Foundatio {
+    public class SharedOptions {
+        public ISerializer Serializer { get; set; }
+        public ILoggerFactory LoggerFactory { get; set; }
+    }
+
+    public interface ISharedOptionsBuilder : IOptionsBuilder {}
```
edwardmeng: I think the `ISharedOptionsBuilder` is not neccessary. Since we can define the extension methods for `IOptionsBuilder` instead.

src/Foundatio/Messaging/SharedMessageBusOptions.cs
niemyjski: I'm not to worried, this was made in 6.0 (a month or two ago) and I doubt anyone besides us has inherited from them.

src/Foundatio/Utility/OptionsBuilder.cs
```diff
@@ -1,17 +1,28 @@
...
-        public static T Build(Action<OptionsBuilder<T>> config) {
-            var builder = new OptionsBuilder<T>();
-            config?.Invoke(builder);
-            return builder.Target;
+        public virtual T Build() {
+            return Target;
         }
     }
+
+    public delegate TBuilder Builder<TBuilder, TOptions>(TBuilder builder) where TBuilder: class, IOptionsBuilder<TOptions>, new();
```
edwardmeng: The delegate name should not so simple that it's too easily confused with mscorlib classes.

src/Foundatio/Queues/SharedQueueOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public class SharedQueueOptions<T> : SharedOptions, ISharedQueueOptions where T : class {
+        public string Name { get; set; } = typeof(T).Name;
+        public int Retries { get; set; } = 2;
+        public TimeSpan WorkItemTimeout { get; set; } = TimeSpan.FromMinutes(5);
+        public IEnumerable<IQueueBehavior<T>> Behaviors { get; set; }
+    }
+
+    public interface ISharedQueueOptionsBuilder : ISharedOptionsBuilder {}
```
edwardmeng: I think the interface `ISharedQueueOptionsBuilder` is not neccessary. We can define the extension methods for `IOptionsBuilder` instead.

src/Foundatio/Utility/SharedOptions.cs
edwardmeng: I am think about whether the `ISharedOptionsBuilder` and `SharedOptionsBuilderExtensions` can be just combined into single class `SharedOptionsBuilder` to reduce the complexity of inherit. I am worrying about it is too complex to create new module and difficult to understand for new comers. For example there should have to 4 classes/interfaces included `ISharedQueueOptions`, `SharedQueueOptions<T>`, `ISharedQueueOptionsBuilder`, `SharedQueueOptionsBuilderExtensions`.

src/Foundatio/Queues/SharedQueueOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    public static class SharedQueueOptionsBuilderExtensions {
```
edwardmeng: There should be extension methods for queue behaviors.

edwardmeng: The generic argument for behaviors is the critical block for the builder design pattern

edwardmeng:
Please refer to my stabs that keep the breaking changes as less as possible
https://github.com/edwardmeng/Foundatio/commit/02e75ee1120ca3a154b0e1a44a1f3278a2f93879

src/Foundatio/Messaging/SharedMessageBusOptions.cs
ejsmith: Yeah, 6.0 is a breaking change. I think Shared is a better word and it didn't need to be an abstract class and that was actually causing issues with some of the fluent builder stuff.

src/Foundatio/Utility/OptionsBuilder.cs
ejsmith: It seems like I had a problem with it being out, but I will try again.

src/Foundatio/Utility/SharedOptions.cs
ejsmith: Yeah, I agree with that. I think it would be better if we just had base options classes with methods and not use the extension methods.

src/Foundatio/Queues/SharedQueueOptions.cs
ejsmith: Yeah, the behavior methods were causing all kinds of problems. But I wonder if it will be better if the methods are on the shared options class directly instead of being extension methods? I will give it a try.

src/Foundatio/Utility/SharedOptions.cs
ejsmith: I just tried to combine into a single options class that you would just derive from, but it doesn't really work because without the methods being extension methods, you can't have them reflect back the same type that they were invoked with.

ejsmith: The actual options classes end up being very simple:

```csharp
public class FolderFileStorageOptionsBuilder : OptionsBuilder<FolderFileStorageOptions>, ISharedOptionsBuilder {
    public FolderFileStorageOptionsBuilder Folder(string folder) {
        if (string.IsNullOrEmpty(folder))
            throw new ArgumentNullException(nameof(folder));
        Target.Folder = folder;
        return this;
    }
}
```

You can bring in the shared options by simply adding the marker interface to your builder and deriving your options class from the sharedoptions. The concrete methods on your builder class end up being very simple too.

src/Foundatio/Utility/OptionsBuilder.cs
ejsmith: It has a generic constraint that makes it so that its not useable with random things. I wanted a name that was simple since we have to pass in both type parameters to it.

src/Foundatio/Queues/SharedQueueOptions.cs
ejsmith: I added the behavior methods to the concrete builder class. It will need to be added to each implementation.

src/Foundatio/Utility/SharedOptions.cs
edwardmeng: Excellent. So the other modules should be improved like this(Queues, Metrics and Messaging)

ejsmith: They should already be the same pattern, are they not? Queues are a bit more complicated since they have generics with the model type.

edwardmeng: I think the `ISharedQueueOptionsBuilder`, `SharedQueueOptions<T>`, `ISharedQueueOptionsBuilder` and `SharedQueueOptionsBuilderExtensions` can be combined to the following code
```cs
    public class SharedQueueOptions<T> : SharedOptions where T : class {
        public string Name { get; set; } = typeof(T).Name;
        public int Retries { get; set; } = 2;
        public TimeSpan WorkItemTimeout { get; set; } = TimeSpan.FromMinutes(5);
        public IEnumerable<IQueueBehavior<T>> Behaviors { get; set; }
    }

    public class SharedQueueOptionsBuilder<T, TOptions, TBuilder> : OptionsBuilder<TOptions> 
        where T : class
        where TOptions : SharedQueueOptions<T>, new()
        where TBuilder : SharedQueueOptionsBuilder<T, TOptions, TBuilder> {
        public TBuilder Name(string name) {
            Target.Name = name;
            return (TBuilder)this;
        }

        public TBuilder Retries(int retries) {
            Target.Retries = retries;
            return (TBuilder)this;
        }

        public TBuilder WorkItemTimeout(TimeSpan timeout) {
            Target.WorkItemTimeout = timeout;
            return (TBuilder)this;
        }

        public TBuilder Behaviors(IEnumerable<IQueueBehavior<T>> behaviors) {
            Target.Behaviors = behaviors;
            return (TBuilder)this;
        }

        public TBuilder AddBehavior(IQueueBehavior<T> behavior) {
            if (behavior == null)
                throw new ArgumentNullException(nameof(behavior));
            if (Target.Behaviors == null)
                Target.Behaviors = new[] { behavior };
            else
                Target.Behaviors = Target.Behaviors.Concat(new[] { behavior });

            return (TBuilder)this;
        }
    }
```
And then the fluent methods for `InMemoryQueueOptions` should be simplified like
```cs
    public class InMemoryQueueOptionsBuilder<T> : SharedQueueOptionsBuilder<T, InMemoryQueueOptions<T>, InMemoryQueueOptionsBuilder<T>> where T: class {
        public InMemoryQueueOptionsBuilder<T> RetryDelay(TimeSpan retryDelay) {
            Target.RetryDelay = retryDelay;
            return this;
        }

        public InMemoryQueueOptionsBuilder<T> RetryMultipliers(int[] multipliers) {
            Target.RetryMultipliers = multipliers ?? throw new ArgumentNullException(nameof(multipliers));
            return this;
        }
    }
```

edwardmeng: The `ISharedOptionsBuilder` and `SharedOptionsBuilderExtensions` should be combined to the following code at first
```cs
    public class SharedOptionsBuilder<TOption, TBuilder> : OptionsBuilder<TOption>
        where TOption : SharedOptions, new()
        where TBuilder : SharedOptionsBuilder<TOption, TBuilder> {
        public TBuilder Serializer(ISerializer serializer) {
            Target.Serializer = serializer ?? throw new ArgumentNullException(nameof(serializer)); ;
            return (TBuilder)this;
        }

        public TBuilder LoggerFactory(ILoggerFactory loggerFactory) {
            Target.LoggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory)); ;
            return (TBuilder)this;
        }
    }
```
And the `SharedQueueOptionsBuilder` should be changed to
```cs
    public class SharedQueueOptionsBuilder<T, TOptions, TBuilder> : SharedOptionsBuilder<TOptions, TBuilder> 
        where T : class
        where TOptions : SharedQueueOptions<T>, new()
        where TBuilder : SharedQueueOptionsBuilder<T, TOptions, TBuilder> {
    }
```

edwardmeng: Please refer to the commit for the full list changes
https://github.com/edwardmeng/Foundatio/commit/27b6196b19fbb8a40bd522c96fb0331d5e2d9673

ejsmith: Yeah, my issue with that is that it forces inheritance from SharedOptions and SharedOptionsBuilder vs right now you can just add the marker interface to your builder and you could still derive from something else. Same thing for the options classes, you could just implement the ISharedOptions (just added) interface and not have to derive from SharedOptions and the builder methods would still work.

The other problem I have is that its making the concrete builder and options classes a lot more verbose with them having to specify both the builder and options types in the inherited classes.

The queue options is really the only one that is kind of breaking things because it uses generics. I tried to figure out a way for it to not use generics, but it seems like it needs them for the behaviors. If you look at the concrete builders and options classes, they are very simple. It's just the shared stuff that is a little verbose / ugly, but that is what we are supposed to be doing. We do the hard stuff and try and make the implementations as straight forward as possible.

Also, I just simplified the behaviors a bit by reducing it down to a single builder method that can take either a single behavior, multiple or an array of them by using a `params` array.

PR #123, opened 2018-03-14
Fix interval between failing job runs

jamie94bc:
The current implementation uses `Milliseconds` rather than `TotalMilliseconds`.

### DIFF
--- a/src/Foundatio.TestHarness/Jobs/HelloWorldJob.cs
+++ b/src/Foundatio.TestHarness/Jobs/HelloWorldJob.cs
@@ -26,6 +26,25 @@ namespace Foundatio.Tests.Jobs {
         }
     }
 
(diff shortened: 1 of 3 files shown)


niemyjski:
Thanks for the PR

niemyjski:
@jamie94bc I merged it and ran the tests with the old code and they pass, what kind of issue did you see with total ms? When we cast it to an int, precision will be lost and roughly the same as the previous implementation.

jamie94bc:
Milliseconds returns the number of milliseconds as a component of the TimeSpan. So `TimeSpan.FromSeconds(1.5).Milliseconds == 500`, not the 1500ms we want to wait for.

niemyjski:
very nice catch!

PR #124, opened 2018-03-14
Trying to handle TaskCanceledException in FolderFileStorage.

edwardmeng:
When the `TaskCanceledException` was thrown in the `SaveFileAsync` method, the log message will be like
```
[18:38:09 ERR] Error trying to save file: q/e46/e46cf1028b974b1d8d217fe9c0fc507d.payload
System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Foundatio.Storage.FolderFileStorage.<SaveFileAsync>d__14.MoveNext() in C:\projects\foundatio\src\Foundatio\Storage\FolderFileStorage.cs:line 98
```
Which cannot identify the root cause of the error.

### DIFF
--- a/src/Foundatio/Storage/FolderFileStorage.cs
+++ b/src/Foundatio/Storage/FolderFileStorage.cs
@@ -98,7 +98,11 @@ namespace Foundatio.Storage {
                     return true;
                 }
             } catch (Exception ex) {
-                _logger.LogError(ex, "Error trying to save file: {Path}", path);
+                const string errorMessage = "Error trying to save file: {Path}";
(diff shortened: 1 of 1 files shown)


niemyjski:
@edwardmeng typically it's because the request was aborted and not awaited. When this happens, ASP.NET core will just abort any operation.

PR #127, opened 2018-05-18
Add ShouldCloneLocalValues to HybridCacheClient

tobyhei:
Allows consumers to configure _localCache.ShouldCloneValues


tobyhei:
We've found this to be quite important for some use cases, e.g when you have a few large objects that are accessed frequently.

niemyjski:
Thanks for the PR! I didn't see a notification for this until just now :\

PR #130, opened 2018-08-15
Add implementation of metrics client with AppMetrics.

edwardmeng:


ejsmith:
This looks good! Thanks!

PR #131, opened 2018-08-20
Downgrade the AppMetrics to 2.0.0 to avoid version conflicts.

edwardmeng:


ejsmith:
Is this just to make it so that people using older versions of appmetrics are able to use this lib?

edwardmeng:
Yes. when we want to add AppMetrics to AspNetCore projects, the latest version App.Metrics.AspNetCore is 2.0.0 which is not compitible to any AppMetrics dependency v2.1.0.  So when the Foundatio.AppMetrics depend on AppMetrics 2.1.0, there will be runtime error thrown. Please refer to the issue: https://github.com/AppMetrics/AppMetrics/issues/297

edwardmeng:
The only way to fix this issue is to downgrade the version of package App.Metrics.Abstraction in Foundatio.AppMetrics.

PR #132, opened 2018-09-18
Publish cache invalidation after distributed cache update

andreaskromann:
Invalidate should happen _after_ the distrubuted cache has been updated. Otherwise it can happen that another cache instance removes the local copy (because of the invalidation) and fetches the old stale value from the distributed cache, because it has not been updated yet.

This actually happens when the get requests are very frequent, and the object serialization takes time or the application is very busy. It can create a delay between invalidate and the distributed cache update in the current version of the code.

### DIFF
--- a/src/Foundatio/Caching/HybridCacheClient.cs
+++ b/src/Foundatio/Caching/HybridCacheClient.cs
@@ -87,15 +87,15 @@ namespace Foundatio.Caching {
         public async Task<int> RemoveAllAsync(IEnumerable<string> keys = null) {
             var items = keys?.ToArray();
             bool flushAll = items == null || items.Length == 0;
-            await _localCache.RemoveAllAsync(items).AnyContext();
             var removed =  await _distributedCache.RemoveAllAsync(items).AnyContext();
+            await _localCache.RemoveAllAsync(items).AnyContext();
             await _messageBus.PublishAsync(new InvalidateCache { CacheId = _cacheId, FlushAll = flushAll, Keys = items }).AnyContext();
             return removed;
         }
 
         public async Task<int> RemoveByPrefixAsync(string prefix) {
-            await _localCache.RemoveByPrefixAsync(prefix).AnyContext();
             var removed = await _distributedCache.RemoveByPrefixAsync(prefix).AnyContext();
+            await _localCache.RemoveByPrefixAsync(prefix).AnyContext();
             await _messageBus.PublishAsync(new InvalidateCache { CacheId = _cacheId, Keys = new[] { prefix + "*" } }).AnyContext();
             return removed;
         }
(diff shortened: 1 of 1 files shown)


niemyjski:
I agree with this. Do you think maybe we should still kick them off (without awaiting) in the old order and then do a Task.WhenAll? cc @ejsmith

andreaskromann:
If you are going to use Task.WhenAll then you still need to keep the invalidation out of it and wait until the others are done. For arguments sake lets say the serialization of the value takes 5 min. Then you can't start sending the invalidation message immediately.

ejsmith:
I think this is fine. If we update the value first then the worst case scenario is that someone gets an old value before it’s invalidated.

src/Foundatio/Caching/HybridCacheClient.cs
```diff
@@ -87,15 +87,17 @@ public class HybridCacheClient : IHybridCacheClient {
         public async Task<int> RemoveAllAsync(IEnumerable<string> keys = null) {
             var items = keys?.ToArray();
             bool flushAll = items == null || items.Length == 0;
-            await _messageBus.PublishAsync(new InvalidateCache { CacheId = _cacheId, FlushAll = flushAll, Keys = items }).AnyContext();
             await _localCache.RemoveAllAsync(items).AnyContext();
-            return await _distributedCache.RemoveAllAsync(items).AnyContext();
+            var removed =  await _distributedCache.RemoveAllAsync(items).AnyContext();
```
niemyjski: I was just giving this one last final review before merging in. Wouldn't we want to invalidate the remote cache first here? Otherwise you have a dirty read and possibly dirty set in the local cache. The same for RemoveByPrefix

src/Foundatio/Caching/HybridCacheClient.cs
```diff
@@ -134,38 +136,43 @@ public class HybridCacheClient : IHybridCacheClient {
...
             await _localCache.ReplaceAsync(key, value, expiresIn).AnyContext();
-            return await _distributedCache.ReplaceAsync(key, value, expiresIn).AnyContext();
+            bool replaced = await _distributedCache.ReplaceAsync(key, value, expiresIn).AnyContext();
+            await _messageBus.PublishAsync(new InvalidateCache { CacheId = _cacheId, Keys = new[] { key } }).AnyContext();
+            return replaced;
         }
 
         public async Task<double> IncrementAsync(string key, double amount, TimeSpan? expiresIn = null) {
-            await _messageBus.PublishAsync(new InvalidateCache { CacheId = _cacheId, Keys = new[] { key } }).AnyContext();
             await _localCache.RemoveAsync(key).AnyContext();
```
niemyjski: We might want to do the remove after the increment, or even better we could take the incremented count and do a replace on the local cache value :)

ejsmith: Yes

andreaskromann:
Hi, I agree with the above two comments. I'll see if I can update the PR sometime this weekend.

andreaskromann:
Hi, I updated the PR with the suggested changes.

niemyjski:
Looks good, do you think we should be removing the local cache last on SetIf operations, notice we did some of them but not all. I'm fine either way.

niemyjski:
Thanks for the pr

PR #136, opened 2018-11-27
Add efficient storage paging

ejsmith:


src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -214,9 +214,9 @@ public FolderFileStorage(Builder<FolderFileStorageOptionsBuilder, FolderFileStor
             
         }
 
-        public Task<IEnumerable<FileSpec>> GetFileListAsync(string searchPattern = null, int? limit = null, int? skip = null, CancellationToken cancellationToken = default) {
+        public async Task<FileListResult> GetFileListAsync(string searchPattern = null, int? limit = null, CancellationToken cancellationToken = default) {
```
niemyjski: Shouldn't this return an interface?

src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -225,9 +225,21 @@ public FolderFileStorage(Builder<FolderFileStorageOptionsBuilder, FolderFileStor
 
             var list = new List<FileSpec>();
             if (!Directory.Exists(Path.GetDirectoryName(Path.Combine(Folder, searchPattern))))
-                return Task.FromResult<IEnumerable<FileSpec>>(list);
+                return FileListResult.Empty;
 
-            foreach (string path in Directory.EnumerateFiles(Folder, searchPattern, SearchOption.AllDirectories).Skip(skip ?? 0).Take(limit ?? Int32.MaxValue)) {
+            var result = new FileListResult(() => Task.FromResult(GetFiles(searchPattern, 1, limit)));
+            await result.NextPageAsync().AnyContext();
```
niemyjski: Do we need to be calling this here, we are already calling GetFiles and populating the initial result.

src/Foundatio/Storage/IFileStorage.cs
```diff
@@ -18,7 +19,61 @@ public interface IFileStorage : IHaveSerializer, IDisposable {
         Task<bool> CopyFileAsync(string path, string targetPath, CancellationToken cancellationToken = default);
         Task<bool> DeleteFileAsync(string path, CancellationToken cancellationToken = default);
         Task DeleteFilesAsync(string searchPattern = null, CancellationToken cancellation = default);
```
niemyjski: Seems like you'd want a result here or a long count on how many were removed

src/Foundatio/Storage/FolderFileStorage.cs
ejsmith: FileSpec isn’t and I ran into issues with trying to make it return an interface.

src/Foundatio/Storage/IFileStorage.cs
ejsmith: Guess we could do that since we are already breaking things.

src/Foundatio/Storage/FolderFileStorage.cs
ejsmith: Yes, it’s not being called yet. It’s passing a lamda that gets the files and it needs to call it the 1st time.

src/Foundatio/Storage/IFileStorage.cs
```diff
@@ -18,7 +19,61 @@ public interface IFileStorage : IHaveSerializer, IDisposable {
...
+    }
+
+    public class FileListResult : IHasNextPageFunc {
+        private static IReadOnlyCollection<FileSpec> _empty = new ReadOnlyCollection<FileSpec>(new FileSpec[0]);
+        public static FileListResult Empty = new FileListResult(_empty);
+
+        public FileListResult(IReadOnlyCollection<FileSpec> files) {
+            Files = files;
+            HasMore = false;
+            ((IHasNextPageFunc)this).NextPageFunc = null;
```
niemyjski: These should be defaulted values.

src/Foundatio/Storage/IFileStorage.cs
```diff
@@ -18,7 +19,61 @@ public interface IFileStorage : IHaveSerializer, IDisposable {
...
+        }
+        
+        public FileListResult(IReadOnlyCollection<FileSpec> files, bool hasMore, Func<Task<NextPageResult>> nextPageFunc) {
+            Files = files;
+            HasMore = hasMore;
+            ((IHasNextPageFunc)this).NextPageFunc = nextPageFunc;
+        }
+
+        public FileListResult(Func<Task<NextPageResult>> nextPageFunc) {
+            ((IHasNextPageFunc)this).NextPageFunc = nextPageFunc;
```
niemyjski: I'm curious what the use cases of this would be? Also how would the current files be populated or has more?

src/Foundatio/Storage/IFileStorage.cs
ejsmith: Huh?

src/Foundatio/Storage/IFileStorage.cs
ejsmith: It’s literally being used in the code you just commented on. You give it a callback and then call nextpage to load the initial page of data. Otherwise you have to repeat yourself.

src/Foundatio/Storage/IFileStorage.cs
```diff
@@ -18,7 +19,61 @@ public interface IFileStorage : IHaveSerializer, IDisposable {
...
-        Task<IEnumerable<FileSpec>> GetFileListAsync(string searchPattern = null, int? limit = null, int? skip = null, CancellationToken cancellationToken = default);
+        Task<FileListResult> GetFileListAsync(string searchPattern = null, int? limit = null, CancellationToken cancellationToken = default);
+    }
+
+    public interface IHasNextPageFunc {
+        Func<Task<NextPageResult>> NextPageFunc { get; set; }
+    }
+
+    public class NextPageResult {
+        public bool Success { get; set; }
```
niemyjski: Would be nice if these were readonly

ejsmith: It’s internal to implementation and I need them to support modifying.

src/Foundatio/Storage/FolderFileStorage.cs
niemyjski: ah, I didn't catch that, was kind of confused for a minute.

src/Foundatio/Storage/IFileStorage.cs
niemyjski: Any reason for setting a value which is already the default value of both HasMore (false) and NextPageFunc (null)

PR #139, opened 2019-03-04
Refactor test time

ejsmith:


src/Foundatio/Utility/Time.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        public void Freeze() {
+            SetFrozenTime(Now());
+        }
+
+        public void Unfreeze() {
+            SetTime(Now());
+        }
+
+        public void SetFrozenTime(DateTime time) {
```
nj: Could be simplified into `public void Freeze(DateTime time) {`, making use of `Freeze` for both scenarious.

src/Foundatio/Utility/Time.cs
```diff
@@ -0,0 +1,224 @@
...
+            SetTime(time, true);
+        }
+
+        public void SetTime(DateTime time, bool freeze = false) {
```
nj: Perhaps splitting into `public void Travel(DateTime time) { SetTime(time, false); }` (inspired by [`timecop`](https://github.com/travisjeffery/timecop)) and changing `private void SetTime(DateTime time, bool freeze = false) { ...`, to better explain what's going on and isolate the internal helper function.

niemyjski: I like this..

PR #145, opened 2019-05-02
Added Message Filter Function in MessageSubscriber

tboeker:
I wrapped up my Messages in a generic "envelope" base message to send additional tracing properties through the message bus. That works fine, but every subscriber connects now to the same message type. aas a result,  there are many many subscribers for each message. the goal of this pull request is to filter the message with an custom function before it will be sent to all subscribers

### DIFF
--- a/src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
+++ b/src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
@@ -10,6 +10,7 @@ using Foundatio.Messaging;
 using Foundatio.Utility;
 using Xunit;
 using Foundatio.AsyncEx;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.Extensions.Logging;
 using Xunit.Abstractions;
 
@@ -192,8 +193,7 @@ namespace Foundatio.Tests.Messaging {
                 await Run.InParallelAsync(iterations, i => messageBus.PublishAsync(new SimpleMessageA { Data = "Hello" }));
                 await countdown.WaitAsync(TimeSpan.FromSeconds(2));
                 Assert.Equal(0, countdown.CurrentCount);
-            }
-            finally {
+            } finally {
                 await CleanupMessageBusAsync(messageBus);
             }
         }
@@ -352,6 +352,35 @@ namespace Foundatio.Tests.Messaging {
             }
         }
 
+        public virtual async Task WillOnlyReceiveSubscribedMessageTypeWithFilterAsync() {
+            var messageBus = GetMessageBus();
+            if (messageBus == null)
+                return;
+
+            try {
+                var countdown = new AsyncCountdownEvent(2);
+
+                await messageBus.SubscribeAsync<SimpleMessageA>(msg => {
+                    Assert.Equal("Hello", msg.Data);
+                    countdown.Signal();
+                });
+
+                await messageBus.SubscribeAsync<SimpleMessageA>(msg => {
+                    Assert.Equal("Hello", msg.Data);
+                    countdown.Signal();
+                }, messagefilter: msg => string.Equals("Hello", msg.Data));
+
+                await messageBus.PublishAsync(new SimpleMessageA {
+                    Data = "Hello"
+                });
+
+                await countdown.WaitAsync(TimeSpan.FromSeconds(2));
+                Assert.Equal(0, countdown.CurrentCount);
+            } finally {
+                await CleanupMessageBusAsync(messageBus);
+            }
+        }
+
         public virtual async Task WillReceiveDerivedMessageTypesAsync() {
             var messageBus = GetMessageBus();
             if (messageBus == null)
@@ -379,6 +408,7 @@ namespace Foundatio.Tests.Messaging {
                 await CleanupMessageBusAsync(messageBus);
             }
         }
+        
 
         public virtual async Task CanSubscribeToAllMessageTypesAsync() {
             var messageBus = GetMessageBus();
@@ -424,7 +454,7 @@ namespace Foundatio.Tests.Messaging {
                     Assert.Equal("Hello", msg.Data);
                     countdown.Signal();
                 });
-                
+
                 await countdown.WaitAsync(TimeSpan.FromMilliseconds(100));
                 Assert.Equal(1, countdown.CurrentCount);
             } finally {
@@ -513,7 +543,7 @@ namespace Foundatio.Tests.Messaging {
             if (messageBus == null)
                 return;
 
-            using (messageBus) {}
+            using (messageBus) { }
         }
     }
 }
--- a/src/Foundatio/Messaging/IMessageSubscriber.cs
+++ b/src/Foundatio/Messaging/IMessageSubscriber.cs
@@ -4,19 +4,25 @@ using System.Threading.Tasks;
 
 namespace Foundatio.Messaging {
     public interface IMessageSubscriber {
-        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, CancellationToken cancellationToken = default) where T : class;
+        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, CancellationToken cancellationToken = default, Func<T, bool> messagefilter = null) where T : class;
+        
     }
 
     public static class MessageBusExtensions {
-        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Func<T, Task> handler, CancellationToken cancellationToken = default) where T : class {
-            return subscriber.SubscribeAsync<T>((msg, token) => handler(msg), cancellationToken);
+        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Func<T, Task> handler, CancellationToken cancellationToken = default, Func<T, bool> messagefilter = null) where T : class {
+            return subscriber.SubscribeAsync<T>((msg, token) => handler(msg), cancellationToken, messagefilter);
         }
 
-        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Action<T> handler, CancellationToken cancellationToken = default) where T : class {
+        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Action<T> handler, CancellationToken cancellationToken = default, Func<T, bool> messagefilter = null) where T : class {
             return subscriber.SubscribeAsync<T>((msg, token) => {
                 handler(msg);
                 return Task.CompletedTask;
-            }, cancellationToken);
+            }, cancellationToken, messagefilter);
         }
+
+
+      
     }
+
+
 }
--- a/src/Foundatio/Messaging/MessageBusBase.cs
+++ b/src/Foundatio/Messaging/MessageBusBase.cs
@@ -38,14 +38,14 @@ namespace Foundatio.Messaging {
             await EnsureTopicCreatedAsync(cancellationToken).AnyContext();
             await PublishImplAsync(GetMappedMessageType(messageType), message, delay, cancellationToken).AnyContext();
         }
- 
+
         private readonly ConcurrentDictionary<Type, string> _mappedMessageTypesCache = new ConcurrentDictionary<Type, string>();
         protected string GetMappedMessageType(Type messageType) {
             return _mappedMessageTypesCache.GetOrAdd(messageType, type => {
                 var reversedMap = _options.MessageTypeMappings.ToDictionary(kvp => kvp.Value, kvp => kvp.Key);
                 if (reversedMap.ContainsKey(type))
                     return reversedMap[type];
-                
+
                 return String.Concat(messageType.FullName, ", ", messageType.Assembly.GetName().Name);
             });
         }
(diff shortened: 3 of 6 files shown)


src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
```diff
@@ -10,6 +10,7 @@
 using Foundatio.Utility;
 using Xunit;
 using Foundatio.AsyncEx;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
```
niemyjski: Is there any reason this is needed?

src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
```diff
@@ -352,6 +352,35 @@ public abstract class MessageBusTestBase : TestWithLoggingBase {
...
+                await messageBus.SubscribeAsync<SimpleMessageA>(msg => {
+                    Assert.Equal("Hello", msg.Data);
+                    countdown.Signal();
+                }, messagefilter: msg => string.Equals("Hello", msg.Data));
+
+                await messageBus.PublishAsync(new SimpleMessageA {
+                    Data = "Hello"
+                });
+
+                await countdown.WaitAsync(TimeSpan.FromSeconds(2));
```
niemyjski: Can we add a test assertion here where the message does not contain hello.

src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
```diff
@@ -379,6 +408,7 @@ public abstract class MessageBusTestBase : TestWithLoggingBase {
                 await CleanupMessageBusAsync(messageBus);
             }
         }
+
```
niemyjski: Some extra whitespace :)

src/Foundatio/Messaging/IMessageSubscriber.cs
```diff
@@ -4,19 +4,25 @@
 
 namespace Foundatio.Messaging {
     public interface IMessageSubscriber {
-        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, CancellationToken cancellationToken = default) where T : class;
+        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, CancellationToken cancellationToken = default, Func<T, bool> messagefilter = null) where T : class;
```
niemyjski: Maybe some discussion around this... How often do we think message filter will be used (also noticed casing on this should be `messageFilter`, or maybe we call this something different like `filter`?)? I kind of like cancellation tokens always being the last argument. Either way this will break the abstraction.

src/Foundatio/Messaging/MessageBusBase.cs
```diff
@@ -80,6 +80,18 @@ public abstract class MessageBusBase<TOptions> : IMessageBus, IDisposable where
                     }
 
                     return handler((T)message, cancellationToken);
+                },
+                MessageFilter = (message) => {
```
niemyjski: Wondering if we should just keep this null by default. I kind of like having the type filtering outside where it was before as it makes it so this func is kind of filtered before we even get to it and it saves serialization overhead.

src/Foundatio/Messaging/MessageBusBase.cs
```diff
@@ -254,8 +269,14 @@ protected class Subscriber {
             public Type Type { get; set; }
             public Func<object, CancellationToken, Task> Action { get; set; }
 
-            public bool IsAssignableFrom(Type type) {
-                return _assignableTypesCache.GetOrAdd(type, t => Type.GetTypeInfo().IsAssignableFrom(t));
+            public Func<object, bool> MessageFilter { get; set; }
+
+            public bool IsAssignableFrom(Type type, object message) {
+
+                if (MessageFilter == null)
```
niemyjski: I guess it's more readable to set this, but I think perf wise when this is called a billion times it would be better to just null check it. Also is there any concern that code in the message filter might throw? I think we should make the assumption that it will never throw and don't put stupid code in there. I also don't think it needs to be async but I can't think of any use case where it would need to be.

src/Foundatio/Messaging/MessageBusBase.cs
```diff
@@ -135,6 +143,13 @@ public abstract class MessageBusBase<TOptions> : IMessageBus, IDisposable where
                 return;
             }
 
+            var subscribers = _subscribers.Values.Where(s => s.IsAssignableFrom(messageType, body)).ToList();
```
niemyjski: I see why this was moved below but this really stinks having to do a serialization hit just to check the body. I'd rather we do the subscriber check above and when it gets to this point check the filter if it's not null.

src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
tboeker: Nope. I copied the code from another project. Maybe Resharper Added this reference.

src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
tboeker: sorry, it's my first pull request ever ;-)

src/Foundatio/Messaging/IMessageSubscriber.cs
tboeker: i don't know. i would use it always ;-)

maybe we can introduce a second method in the IMessageSubscriber interface like this:

```

 public interface IMessageSubscriber {


        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, CancellationToken cancellationToken = default) where T : class;

        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, Func<T, bool> filter, CancellationToken cancellationToken = default) where T : class;

    }

 public static class MessageBusExtensions {
        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Func<T, Task> handler, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => handler(msg), cancellationToken);
        }

        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Action<T> handler, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => {
                handler(msg);
                return Task.CompletedTask;
            }, cancellationToken);
        }


        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Func<T, Task> handler, Func<T, bool> filter, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => handler(msg), cancellationToken, filter);
        }

        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Action<T> handler, Func<T, bool> filter, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => {
                handler(msg);
                return Task.CompletedTask;
            }, cancellationToken, filter);
        }


    }

```

tboeker: we could change it this way:

```
   public interface IMessageSubscriber {


        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, CancellationToken cancellationToken = default) where T : class;

        Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, Func<T, bool> filter, CancellationToken cancellationToken = default) where T : class;

    }

    public static class MessageBusExtensions {
        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Func<T, Task> handler, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => handler(msg), cancellationToken);
        }

        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Action<T> handler, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => {
                handler(msg);
                return Task.CompletedTask;
            }, cancellationToken);
        }

        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Func<T, Task> handler, Func<T, bool> filter, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => handler(msg), filter, cancellationToken);
        }

        public static Task SubscribeAsync<T>(this IMessageSubscriber subscriber, Action<T> handler, Func<T, bool> filter, CancellationToken cancellationToken = default) where T : class {
            return subscriber.SubscribeAsync<T>((msg, token) => {
                handler(msg);
                return Task.CompletedTask;
            }, filter, cancellationToken);
        }
        
    }
```

no breaking changes. only optional parameter

src/Foundatio/Messaging/MessageBusBase.cs
tboeker: You're right. only the number of real subscribers could be different. but i think it would be ok.

src/Foundatio/Messaging/MessageBusBase.cs
tboeker:
```suggestion
                  if (MessageFilter == null)
                    return _assignableTypesCache.GetOrAdd(type, t => Type.GetTypeInfo().IsAssignableFrom(t));

                return _assignableTypesCache.GetOrAdd(type, t => Type.GetTypeInfo().IsAssignableFrom(t) && MessageFilter(message));
```
The MessageFilter is not async. It is a Function

src/Foundatio.TestHarness/Messaging/MessageBusTestBase.cs
niemyjski: No problem (it's not a problem, I need to update our style settings so the editor just does it for you, it's our issue not yours). Thanks again for making a pr, we appreciate it :).

src/Foundatio/Messaging/MessageBusBase.cs
niemyjski: Yeah, I know that. I'm just thinking out loud if there is ever a valid use case where a filter would need to be async.

src/Foundatio/Messaging/MessageBusBase.cs
tboeker: when we introduce the second interface method we can do it like that

```suggestion
                 public async Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, Func<T, bool> filter, CancellationToken cancellationToken = default) where T : class {
            if (_logger.IsEnabled(LogLevel.Trace))
                _logger.LogTrace("Adding subscriber for {MessageType}.", typeof(T).FullName);
            await EnsureTopicSubscriptionAsync(cancellationToken).AnyContext();
            await SubscribeImplAsync(handler, cancellationToken, filter).AnyContext();
        }

        public async Task SubscribeAsync<T>(Func<T, CancellationToken, Task> handler, CancellationToken cancellationToken = default) where T : class {
            if (_logger.IsEnabled(LogLevel.Trace))
                _logger.LogTrace("Adding subscriber for {MessageType}.", typeof(T).FullName);
            await EnsureTopicSubscriptionAsync(cancellationToken).AnyContext();
            await SubscribeImplAsync(handler, cancellationToken, msg => true).AnyContext();
        }
        
        ,
                MessageFilter = (message) => {
               
                    if (!(message is T)) {
                        if (_logger.IsEnabled(LogLevel.Trace))
                            _logger.LogTrace("Unable to call subscriber messagefilter: {MessageType} cannot be safely casted to {SubscriberType}", message.GetType(), typeof(T));
                        return true;
                    }

                    return filter((T)message);
                }
        
```

maybe we can integrate the filter into the subscriber action

src/Foundatio/Messaging/IMessageSubscriber.cs
ejsmith: I want to change subscribe to have an options object so that we can do other things on it as well. Like give it a subscription name or whether it’s durable etc. The message filter can be one of the options.

src/Foundatio/Messaging/MessageBusBase.cs
tboeker: :-)
you should better do the async action in the message handler. not in the filter

src/Foundatio/Messaging/MessageBusBase.cs
niemyjski: I'd rather have one interface and method. I wonder if we can put this on the options object. There is a unified messaging branch that I believe has this options class, perhaps we can introduce it here and build off it in the other branch? @ejsmith

ejsmith:
@tboeker I think this is a good idea and a generally useful feature. But I want to do something a little more flexible so we can add additional options to message subscriptions. I'd like to create some sort of generic message options object that gets passed into the SubscribeAsync method. I'm actually thinking that we should remove the generic type parameter from that method and just add a MessageType to the options object. So you wouldn't have to use generics to create a subscription if the type you are subscribing is not known at compile time. So I think the base level options would have MessageType and your message filter function.

Question is, how would we make it so that any implementation can have their own options and make it easy to set those for end users. The big problem here is that whatever we do here really should be restricted to what makes sense for all messaging implementations otherwise we will be leaking implementation throughout the apps code and that is what we are trying to avoid. I do think it would probably be OK though if we had some implementation specific options that you could set and they would just be ignored in other implementations. For example, if we have a IsDurable option that would just be ignored in the in memory messaging implementation.

tboeker:
i gave up this approach with the generic messages and turned back to strictly typed messages.

ejsmith:
@tboeker ok, glad you have a workaround for now. I have done a bunch of work in the unified-messaging branch and we would be able to do what you want with those changes. I just need to find the time to finish that work.

Issue #239, opened 2020-07-27
[RedisQueue] StartWorkingImpl worker loop throws unhandled exception in task which prevents further queue processing


Seadoo2017:
Consider the following scenario
1. StartWorkingImpl loop: Item is dequeued and user handle function is called with the queue entry
2. User handling function takes more time than queue WorkItemTimeout so the queue entry is "auto abandoned" (while in processing)
3. User handling function finish and calls CompleteAsync on the queue entry
4. RedisQueue::CompleteAsync(IQueueEntry<T> entry) will throw exception (as expected) here:

            long result = await Run.WithRetriesAsync(() => Database.ListRemoveAsync(_workListName, entry.Id), logger: 
            _logger).AnyContext();
            if (result == 0)
                throw new InvalidOperationException("Queue entry not in work list, it may have been auto abandoned.");

which will be handled here:

                     try
                    {
                        await handler(queueEntry, linkedCancellationToken.Token).AnyContext();
                        if (autoComplete && !queueEntry.IsAbandoned && !queueEntry.IsCompleted)
                            await queueEntry.CompleteAsync().AnyContext();
                    }
                    catch (Exception ex)
                    {
                            Interlocked.Increment(ref _workerErrorCount);
                            if (_logger.IsEnabled(LogLevel.Error))
                                _logger.LogError(ex, "Worker error: {Message}", ex.Message);
                            if (!queueEntry.IsAbandoned && !queueEntry.IsCompleted)
                                await queueEntry.AbandonAsync().AnyContext();
                    }

5. Then, inside the catch, the following condition will be true: (the work item is the original non abandoned item which we got when the handler was called)

        if (!queueEntry.IsAbandoned && !queueEntry.IsCompleted)
                   await queueEntry.AbandonAsync().AnyContext();

6. The call to **AbandonAsync** is going to throw the unhandled exception and kill the task loop, because the following transaction will fail (the item was auto abandoned...):

                var tx = Database.CreateTransaction();
                tx.AddCondition(Condition.KeyExists(GetRenewedTimeKey(entry.Id)));
                tx.ListRemoveAsync(_workListName, entry.Id);
                //tx.ListLeftPushAsync(_queueListName, entry.Id);
                tx.SortedSetAddAsync(_queueListName, entry.Id, entry.Value.Priority);
                tx.KeyDeleteAsync(GetRenewedTimeKey(entry.Id));
[...]

PR #240, opened 2020-07-28
Add unit test which reproduce issue #239

Seadoo2017:


niemyjski:
@ejsmith do we have test coverage for this now?

ejsmith:
Yes, it was added.

PR #259, opened 2021-01-31
Use async serialization with System.Text.Json

benaadams:
Rather than going via intermediary  `StreamReader` or `Utf8JsonWriter`


niemyjski (approved):

niemyjski:
Thanks for the PR!

PR #260, opened 2021-02-01
Add async operations to serializer. Add ability to get read or write streams from IFileStorage

ejsmith:


ejsmith:
@benaadams what do you think about these changes?

benaadams:
I wasn't sure what changing the interface would do compat wise and if that would be a problem :)

For the file interface is there a problem with some of the file providers as they have upload blob type implementations rather than stream based?

ejsmith:
Generally, I just try to keep source compat which we can do with extension methods.

Yeah, I think we can make the blob based file providers that require a single PUT operations work one way or another. Worst case scenario we can return a memorystream that on dispose does the blob PUT operation.

PR #262, opened 2021-06-18
Propagate exceptions raised by subscriber message handlers

jcono:
When using the AzureServiceBus the best way to guarantee the delivery of messages is to use `ReceiveMode.PeekLock` on the subscription client.  For details see FoundatioFx/Foundatio.AzureServiceBus#38.

In order for this to work correctly the handler of a message needs to be able to throw an exception up to the subscription client.  These changes allow that to happen by optionally running the handlers sequentially rather than in parallel in a task as they are by default so that if a handler throws an exception it can be handled by the underlying client.

### DIFF
--- a/src/Foundatio/Messaging/InMemoryMessageBus.cs
+++ b/src/Foundatio/Messaging/InMemoryMessageBus.cs
@@ -56,8 +56,9 @@ namespace Foundatio.Messaging {
 
             try {
                 await SendMessageToSubscribers(messageData);
-            } catch {
+            } catch (Exception ex) {
                 // swallow exceptions from subscriber handlers for the in memory bus
+                _logger.LogWarning(ex, "Error sending message to subscribers: {ErrorMessage}", ex.Message);
             }
         }
     }
--- a/src/Foundatio/Messaging/InMemoryMessageBus.cs
+++ b/src/Foundatio/Messaging/InMemoryMessageBus.cs
@@ -31,20 +31,20 @@ namespace Foundatio.Messaging {
             _messageCounts.Clear();
         }
 
-        protected override Task PublishImplAsync(string messageType, object message, TimeSpan? delay, CancellationToken cancellationToken) {
+        protected override async Task PublishImplAsync(string messageType, object message, TimeSpan? delay, CancellationToken cancellationToken) {
             Interlocked.Increment(ref _messagesSent);
             _messageCounts.AddOrUpdate(messageType, t => 1, (t, c) => c + 1);
             Type mappedType = GetMappedMessageType(messageType);
 
             if (_subscribers.IsEmpty)
-                return Task.CompletedTask;
+                return;
 
             bool isTraceLogLevelEnabled = _logger.IsEnabled(LogLevel.Trace);
             if (delay.HasValue && delay.Value > TimeSpan.Zero) {
                 if (isTraceLogLevelEnabled)
                     _logger.LogTrace("Schedule delayed message: {MessageType} ({Delay}ms)", messageType, delay.Value.TotalMilliseconds);
                 SendDelayedMessage(mappedType, message, delay.Value);
-                return Task.CompletedTask;
+                return;
             }
             
             var body = SerializeMessageBody(messageType, message);
@@ -55,12 +55,10 @@ namespace Foundatio.Messaging {
             };
 
             try {
-                SendMessageToSubscribers(messageData);
+                await SendMessageToSubscribers(messageData);
             } catch {
                 // swallow exceptions from subscriber handlers for the in memory bus
             }
-
-            return Task.CompletedTask;
         }
     }
 }
--- a/src/Foundatio/Messaging/InMemoryMessageBus.cs
+++ b/src/Foundatio/Messaging/InMemoryMessageBus.cs
@@ -31,31 +31,31 @@ namespace Foundatio.Messaging {
             _messageCounts.Clear();
         }
 
-        protected override Task PublishImplAsync(string messageType, object message, TimeSpan? delay, CancellationToken cancellationToken) {
+        protected override async Task PublishImplAsync(string messageType, object message, TimeSpan? delay, CancellationToken cancellationToken) {
             Interlocked.Increment(ref _messagesSent);
             _messageCounts.AddOrUpdate(messageType, t => 1, (t, c) => c + 1);
-            Type mappedType = GetMappedMessageType(messageType);
+            var mappedType = GetMappedMessageType(messageType);
 
             if (_subscribers.IsEmpty)
-                return Task.CompletedTask;
+                return;
 
             bool isTraceLogLevelEnabled = _logger.IsEnabled(LogLevel.Trace);
             if (delay.HasValue && delay.Value > TimeSpan.Zero) {
                 if (isTraceLogLevelEnabled)
                     _logger.LogTrace("Schedule delayed message: {MessageType} ({Delay}ms)", messageType, delay.Value.TotalMilliseconds);
                 SendDelayedMessage(mappedType, message, delay.Value);
-                return Task.CompletedTask;
+                return;
             }
-            
-            var body = SerializeMessageBody(messageType, message);
+
+            byte[] body = SerializeMessageBody(messageType, message);
             var messageData = new Message(() => DeserializeMessageBody(messageType, body)) {
                 Type = messageType,
                 ClrType = mappedType,
                 Data = body
             };
 
-            SendMessageToSubscribers(messageData);
-            return Task.CompletedTask;
+            await SendMessageToSubscribers(messageData);
+            return;
         }
     }
 }
--- a/src/Foundatio/Messaging/InMemoryMessageBus.cs
+++ b/src/Foundatio/Messaging/InMemoryMessageBus.cs
@@ -46,7 +46,7 @@ namespace Foundatio.Messaging {
                 SendDelayedMessage(mappedType, message, delay.Value);
                 return Task.CompletedTask;
             }
-
+            
             var body = SerializeMessageBody(messageType, message);
             var messageData = new Message(() => DeserializeMessageBody(messageType, body)) {
                 Type = messageType,
--- a/src/Foundatio/Messaging/InMemoryMessageBus.cs
+++ b/src/Foundatio/Messaging/InMemoryMessageBus.cs
@@ -57,7 +57,7 @@ namespace Foundatio.Messaging {
             try {
                 SendMessageToSubscribers(messageData);
             } catch {
-                // swallow exceptions from subscribers
+                // swallow exceptions from subscriber handlers for the in memory bus
             }
 
             return Task.CompletedTask;
--- a/src/Foundatio/Messaging/InMemoryMessageBus.cs
+++ b/src/Foundatio/Messaging/InMemoryMessageBus.cs
@@ -55,7 +55,6 @@ namespace Foundatio.Messaging {
             };
 
             await SendMessageToSubscribers(messageData);
-            return;
         }
     }
 }
--- a/src/Foundatio/Messaging/InMemoryMessageBus.cs
+++ b/src/Foundatio/Messaging/InMemoryMessageBus.cs
@@ -34,7 +34,7 @@ namespace Foundatio.Messaging {
         protected override Task PublishImplAsync(string messageType, object message, TimeSpan? delay, CancellationToken cancellationToken) {
             Interlocked.Increment(ref _messagesSent);
             _messageCounts.AddOrUpdate(messageType, t => 1, (t, c) => c + 1);
-            var mappedType = GetMappedMessageType(messageType);
+            Type mappedType = GetMappedMessageType(messageType);
 
             if (_subscribers.IsEmpty)
                 return Task.CompletedTask;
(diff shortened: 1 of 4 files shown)


src/Foundatio/Messaging/SharedMessageBusOptions.cs
```diff
@@ -12,6 +12,11 @@ public class SharedMessageBusOptions : SharedOptions {
         /// Controls which types messages are mapped to.
         /// </summary>
         public Dictionary<string, Type> MessageTypeMappings { get; set; } = new Dictionary<string, Type>();
+
+        /// <summary>
+        /// Controls the way the messages are sent to subscribers
+        /// </summary>
+        public bool SendMessagesSynchronously { get; set; } = false;
```
ejsmith, line 19: I really don't like this being an option. As long as we make sure the overall process happens in a separate thread and not in the UI thread then we can probably always run the message handlers synchronously. Most of the messaging implementations already receive messages in a new thread, but I think this was an issue in the in memory message bus implementation. So maybe we can just change that to call the overall thing in a new thread and then make the change to run the message handlers sync.

ejsmith (changes requested):

src/Foundatio/Messaging/InMemoryMessageBus.cs
```diff
@@ -31,31 +31,30 @@ public InMemoryMessageBus(Builder<InMemoryMessageBusOptionsBuilder, InMemoryMess
...
             var body = SerializeMessageBody(messageType, message);
             var messageData = new Message(() => DeserializeMessageBody(messageType, body)) {
                 Type = messageType,
                 ClrType = mappedType,
                 Data = body
             };
 
-            SendMessageToSubscribers(messageData);
-            return Task.CompletedTask;
+            await SendMessageToSubscribers(messageData);
```
niemyjski, line 57: We could also just return this task and the whole method doesn't need to be marked async.

src/Foundatio/Messaging/SharedMessageBusOptions.cs
ejsmith: Also, I remember more about why we did it this way now. We didn't want a single bad subscriber to kill delivery of the message to all the other subscribers.

ejsmith: I'm not sure how other messaging libraries handle this. There could be 10 subscribers and only one of them is throwing an error. Should the one that throws an error stop the entire message delivery to the 9 other subscribers.

niemyjski: I would think the subscriber would handle the error and the other 9 would also get the message.

src/Foundatio/Messaging/MessageBusBase.cs
```diff
@@ -167,6 +167,8 @@ public abstract class MessageBusBase<TOptions> : IMessageBus, IDisposable where
                 return;
             }
 
+            static void LaunchTask(Func<Task> task) => Task.Factory.StartNew(task);
```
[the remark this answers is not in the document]
jcono: Only to get around the fact that there is a warning when you don't `await` the call to `Task.Factory.StartNew` in what is now an async method.

src/Foundatio/Messaging/SharedMessageBusOptions.cs
jcono: Yeah I figured there might be a reason not to do it this way but it solve my problem for now.

Perhaps there's a way to override behaviour in the AzureServiceBus implementation itself?

jcono: I have to apologies as I'm not as familiar with this codebase as I'd like to be to have this conversation but maybe the option is poorly named.  What I'm essentially after is *guaranteed message delivery*.

That is basically the scenario I'm trying to solve for.  I want the publisher to hold on to the message until it is sure that all the subscribers have successfully handled it.  If they don't succeed then the message doesn't get lost and can be delivered later.

So in some ways (if there were multiple subscribers) then I might want a single subscriber that throws an exception to result in the message (eventually) being resent to the subscribers.  If 1 fails they all get the message again.

As far as I can tell the current implementation is more *fire and forget* in that it assumes once you send a message to a subscriber that is good enough, which it probably is in a lot of scenarios.

Also maybe redelivery of the message is possibly specific to the implementation (AzureServiceBus) I'm using but I still thought it might have a place more broadly?

src/Foundatio/Messaging/InMemoryMessageBus.cs
jcono: Fair enough

jcono:
I've made some modifications that I think might be less of a change to the original code.

I've removed the option which I agree wasn't great.  The only real difference now from the original code is that the call to `SendMessageToSubscribers` will wait for *all* the subscriber handlers to finish and will throw an exception if one or more of them don't succeed.

So this handles the problem that one "bad" subscriber can stop message delivery to all the others while also running them in parallel.

The only thing I'm a bit unsure about is whether it's a problem to throw an exception now that was being swallowed before?

ejsmith:
I think this is better now. Can you clean up the whitespace and other minor changes so we can keep this PR small and focused.

jcono:
> I think this is better now. Can you clean up the whitespace and other minor changes so we can keep this PR small and focused.

Sure thing.  Changes are as small as I could make them now I think.

jcono:
I have to apologise, but a bit further testing showed that, for reasons I haven't fully got to the bottom of yet, the way the task was created that was being waited on when sending messages wasn't propagating any exception correctly.  Something about the way the `Task` was being handled.

I've refactored that part of the code and it's now working as I expected.

ejsmith:
This is looking good, but I'm concerned about using Task.WaitAll and it not being async. Seems like that is going to be a sync over async call which is dangerous. Can't we just do `await Task.WhenAll`? Also, I'm a little concerned about doing a Task.Run inside a loop. If there are a lot of subscribers it could cause an issue with thread starvation. That is what we were doing before so that is an existing issue. Could we do a Parallel.ForEach?

ejsmith:
Also, I'd like for us to add a test that shows that multiple good subscribers and one bad one will work as expected and call all of the subscribers even though the one is going to blow up. Maybe register 2 good ones and then 1 bad one and then another 2 good ones and then verify that all of the good ones are called and that the error message (aggregateexception) is bubbled up like you are wanting it to be.

jcono:
> concerned about using Task.WaitAll and it not being async. Seems like that is going to be a sync over async call which is dangerous. Can't we just do `await Task.WhenAll`?

I'd tend to agree, although that will change the entire `SendMessageToSubscribers` method to an async method which is a bigger change (one I had originally).  I'm happy to do that but it will require changing a few more files (and I believe is a possibly breaking change for any libraries as it's a `protected` method)

jcono:
> Also, I'd like for us to add a test that shows that multiple good subscribers and one bad one will work as expected 

I've added to the already existing test `CanTolerateSubscriberFailureAsync()` in the test harness which proves the subscribers still receive the message.  

Confirming the exception contains the appropriate message is a little harder to put into the base functionality as it's really an implementation thing I think.  For example, to pass that test I had to change the `InMemoryMessageBus` to swallow the exception in its publish implementation for it to work (and keep it simple).

Thoughts?

ejsmith:
> > concerned about using Task.WaitAll and it not being async. Seems like that is going to be a sync over async call which is dangerous. Can't we just do `await Task.WhenAll`?
> 
> I'd tend to agree, although that will change the entire `SendMessageToSubscribers` method to an async method which is a bigger change (one I had originally). I'm happy to do that but it will require changing a few more files (and I believe is a possibly breaking change for any libraries as it's a `protected` method)

Yeah, I think we need to make that change because doing sync over async could potentially result in deadlocks.

ejsmith:
> > Also, I'd like for us to add a test that shows that multiple good subscribers and one bad one will work as expected
> 
> I've added to the already existing test `CanTolerateSubscriberFailureAsync()` in the test harness which proves the subscribers still receive the message.
> 
> Confirming the exception contains the appropriate message is a little harder to put into the base functionality as it's really an implementation thing I think. For example, to pass that test I had to change the `InMemoryMessageBus` to swallow the exception in its publish implementation for it to work (and keep it simple).
> 
> Thoughts?

Ahh, I see it now. Yes, that is good. Thanks!

jcono:
Ok all done I think.  The build failed but I'm not sure it's anything to do with the code I've changed so hopefully it'll be ok this time.

jcono (commented):
I think this is all done now?

ejsmith:
Sorry, I'm out on vacation. I think the changes are good, but we just need to verify that we didn't break anything. I'll be out for another couple weeks. Maybe @niemyjski can take a look and see what he thinks.

src/Foundatio/Messaging/InMemoryMessageBus.cs
```diff
@@ -54,8 +54,11 @@ public InMemoryMessageBus(Builder<InMemoryMessageBusOptionsBuilder, InMemoryMess
                 Data = body
             };
 
-            SendMessageToSubscribers(messageData);
-            return Task.CompletedTask;
+            try {
+                await SendMessageToSubscribers(messageData);
+            } catch {
+                // swallow exceptions from subscriber handlers for the in memory bus
```
niemyjski, line 60: I think I'd rather have the logging up here because then we catch all exceptions thrown from `SendMessageToSubscribers` instead of only the ones we awaited for subscription handlers, especially since if the code changes some day and a handler is removed we could be missing out on knowing that a collection was mutated.

src/Foundatio/Messaging/MessageBusBase.cs
```diff
@@ -190,19 +190,23 @@ public abstract class MessageBusBase<TOptions> : IMessageBus, IDisposable where
...
                 });
+            });
+
+            try {
+                await Task.WhenAll(subscriberHandlers.ToArray());
+            } catch (Exception ex) {
+                if (_logger.IsEnabled(LogLevel.Warning))
+                    _logger.LogWarning(ex, "Error sending message to subscribers: {ErrorMessage}", ex.Message);
+
+                throw;
```
niemyjski, line 209: Can we move this up, and then just return the Task.WhenAll and keep this async. This would ensure we log all exceptions inside of SendMessageToSubscribers

niemyjski (changes requested):
The changes look good to me, I'm just concerned about not seeing some bugs due to not logging on the outside of the when all and just eating exceptions. Can we move this logging up a little and then we'll merge it and do some additional testing. Thanks again for the PR.

src/Foundatio/Messaging/MessageBusBase.cs
jcono: Not sure I follow what you mean here.

As far as I can tell, the logging now is almost the same as it was before the change except for the fact that instead of getting a distinct log message for each subscriber handler action that throws an exception there will be a single log message aggregating (AggregateException) all of the exceptions.  Also because the aggregate exception is thrown the trace log message at the bottom doesn't happen either which it would have before.

The swallowing of the exceptions is really just an implementation concern of `InMemoryMessageBus` I think.  Other implementations would be unlikely to be sending messages in the same method that publishes them as they'd be in separate processes (i.e. not in memory) wouldn't they?

Perhaps inside the `InMemoryMessageBus` exception handler that swallows the exception we should just be logging a message as well?

niemyjski: I'm just thinking that we could be eating exceptions in this method and only capturing the exceptions the tasks themselves throw. Moving the logging up to my previous comment located to the empty catch block here -> `await SendMessageToSubscribers(messageData);` would ensure all the logic to setting up the tasks would be covered by logging (lines 153-179).

> The swallowing of the exceptions is really just an implementation concern of InMemoryMessageBus I think. Other implementations would be unlikely to be sending messages in the same method that publishes them as they'd be in separate processes (i.e. not in memory) wouldn't they?

Generally yes.

jcono: Right I think I see now. To clarify, if the call to `SendMessageToSubscribers` in `InMemoryMessageBus` fails for any reason other than a subscriber handler exception we will both swallow it and there'll be no log message?

In which case would it be best to keep the logging in the base class as it is and have the `InMemoryMessageBus` log the exception?  Again, having not used it, and without knowing too much about the use case of the `InMemoryMessageBus` it does seem like an implementation specific concern.  I'd have thought it'd be better to let the exception propagate under normal circumstances (i.e. not swallow it at all) but in order to pass the tests I had to do something with it to allow the `CanTolerateSubscriberFailureAsync` to pass for the `InMemoryMessageBus`.

Anyway I'll put a log message in and see what you think.

niemyjski: Sorry for all the back and forth on this, we'll get this merged ASAP. I was wondering this morning should we be eating exceptions when you call publish in the in memory message bus. Seems like you'd want to know if an error occurred during publish, perhaps log and rethrow or just don't even try catch it? Guess it's nice to have logging.

ejsmith (approved):
Thank you for your efforts on this @jcono. Sorry it took so long.

PR #267, opened 2021-11-10
Updated DeepCloner to 0.10.3

niemyjski:
https://github.com/force-net/DeepCloner/releases/tag/v0.10.3


niemyjski:
@ejsmith Noticed some extra locking added to this release, wanted to bring it to attention in case we notice any issues.

PR #268, opened 2021-12-13
Support for UniqueId on QueueEntryOptions (with MessageBus support)

jcono:
Driven by the need to have duplicate detection functionality Azure Service Bus (see FoundatioFx/Foundatio.AzureServiceBus#39).

Adds a `UniqueId` property to the `QueueEntryOptions` class and supports passing options publishing a message to an `IMessageBus`.

### DIFF
--- a/src/Foundatio/Messaging/IMessageBus.cs
+++ b/src/Foundatio/Messaging/IMessageBus.cs
@@ -2,4 +2,9 @@
 
 namespace Foundatio.Messaging {
     public interface IMessageBus : IMessagePublisher, IMessageSubscriber, IDisposable {}
+    
+    public class MessageOptions {
+        public string UniqueId { get; set; }
+    }
+
 }
--- a/src/Foundatio/Messaging/IMessageBus.cs
+++ b/src/Foundatio/Messaging/IMessageBus.cs
@@ -1,10 +1,12 @@
 ﻿using System;
+using Foundatio.Utility;
 
 namespace Foundatio.Messaging {
     public interface IMessageBus : IMessagePublisher, IMessageSubscriber, IDisposable {}
     
     public class MessageOptions {
         public string UniqueId { get; set; }
+        public string CorrelationId { get; set; }
+        public DataDictionary Properties { get; set; } = new DataDictionary();
     }
-
 }
--- a/src/Foundatio/Messaging/IMessagePublisher.cs
+++ b/src/Foundatio/Messaging/IMessagePublisher.cs
@@ -1,10 +1,11 @@
 ﻿using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Foundatio.Queues;
 
 namespace Foundatio.Messaging {
     public interface IMessagePublisher {
-        Task PublishAsync(Type messageType, object message, TimeSpan? delay = null, CancellationToken cancellationToken = default);
+        Task PublishAsync(Type messageType, object message, TimeSpan? delay = null, QueueEntryOptions options = null, CancellationToken cancellationToken = default);
     }
 
     public static class MessagePublisherExtensions {
--- a/src/Foundatio/Messaging/IMessagePublisher.cs
+++ b/src/Foundatio/Messaging/IMessagePublisher.cs
@@ -5,12 +5,12 @@ using Foundatio.Queues;
 
 namespace Foundatio.Messaging {
     public interface IMessagePublisher {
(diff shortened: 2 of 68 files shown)


src/Foundatio/Messaging/IMessagePublisher.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Foundatio.Queues;
 
 namespace Foundatio.Messaging {
     public interface IMessagePublisher {
-        Task PublishAsync(Type messageType, object message, TimeSpan? delay = null, CancellationToken cancellationToken = default);
+        Task PublishAsync(Type messageType, object message, TimeSpan? delay = null, QueueEntryOptions options = null, CancellationToken cancellationToken = default);
```
niemyjski, line 8: Why are we passing in a QueueEntryOptions into a messaging implementation?

ejsmith: Yeah, this is a breaking change. We can't really change this.

jcono: I did wonder about this.  Does it need a `MessageEntryOptions` instead then?

jcono: To be clear, either there is some sort of breaking change or we go back to having the message object itself implement an known interface in order to provide the unique message id (but you guys suggested otherwise in FoundatioFx/Foundatio.AzureServiceBus#39 unless I misunderstood your intent).  

Unless you can think of another option?

jcono: @niemyjski and @ejsmith 

Any steer on what you would like to do with this?

ejsmith: I thought you were just trying to make this change to the queue abstraction which can be made without breaking changes. For the `IMessagePublisher` we would have to do it some other way. Perhaps your approach of the interface would be OK for now. The long term plan is that we need to consolidate queuing and messaging into a single abstraction and when we do that they would both be able to support metadata. Do you need this feature for messaging right now as well as queuing?

jcono: Yeah I primarily need it for messaging but I do use both.  I've had to essentially fork your code to be able to use it for our purposes (with the duplication) but I was hoping it was useful enough to incorporate into the packages and consume them that way so I didn't diverge over time and create a maintenance issue.

Happy to take your steer on what to do?  Go back to the interface?  Wait till you've had time to consolidate?

ejsmith: I am hoping to get time to work on the consolidation soon, but I think it the meantime, if you wanted to submit a PR that makes the `IMessagePublisher` more like the queue one where we have `MessageOptions` like you suggested as a parameter similar to how we have `QueueEntryOptions` on queues now. That way we aren't adding a parameter to the contract for a single feature.

jcono: Ok thanks.  I'll try and get that together over the next few days (time permitting).

jcono: @ejsmith I've updated this PR (and the ones in the other repositories) with a `MessageOptions` instead.  Let me know what you guys think.

src/Foundatio/Messaging/IMessageBus.cs
```diff
@@ -2,4 +2,9 @@
 
 namespace Foundatio.Messaging {
     public interface IMessageBus : IMessagePublisher, IMessageSubscriber, IDisposable {}
+    
+    public class MessageOptions {
+        public string UniqueId { get; set; }
```
ejsmith, line 7: Let's add a CorrelationId and Properties on here like QueueEntryOptions has.

src/Foundatio/Messaging/IMessagePublisher.cs
ejsmith: It's looking good. Thanks for your efforts with this!

src/Foundatio/Messaging/IMessageBus.cs
jcono: Done, although I'm not sure if I need to use them in any of the implementations.

ejsmith: That's fine. We should be able to support them in a few of the systems. We can add that implementation.

ejsmith (approved):
LGTM

PR #284, opened 2023-01-02
Send work item failed progress messages

### LINKED ISSUE #76: Send work item failed progress messages
We don't send a progress message when a work item fails. What should the status be?

akshay-zz:
Fix #76

### DIFF
--- a/src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
+++ b/src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
@@ -132,6 +132,9 @@ namespace Foundatio.Jobs {
 
                 return JobResult.Success;
             } catch (Exception ex) {
+
+                await ReportProgressAsync(handler, queueEntry, -1, $"Failed: {ex.Message}").AnyContext();
+
                 if (!queueEntry.IsAbandoned && !queueEntry.IsCompleted) {
                     await queueEntry.AbandonAsync().AnyContext();
                     return JobResult.FromException(ex, $"Abandoning {queueEntry.Value.Type} work item: {queueEntry.Id}: Error in handler {workItemDataType.Name}");
@@ -186,7 +189,7 @@ namespace Foundatio.Jobs {
                         string[] typeParts = type.Split(',');
                         if (typeParts.Length >= 2)
                             type = String.Join(",", typeParts[0], typeParts[1]);
-                        
+
                         // try resolve type without version
                         return Type.GetType(type);
                     } catch (Exception ex) {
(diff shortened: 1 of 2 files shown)


tests/Foundatio.Tests/Jobs/WorkItemJobTests.cs
```diff
@@ -246,7 +246,7 @@ public class WorkItemJobTests : TestWithLoggingBase {
 
             await job.RunUntilEmptyAsync();
             await countdown.WaitAsync(TimeSpan.FromSeconds(2));
-            Assert.Equal(0, countdown.CurrentCount);
```
niemyjski, line 249: We should keep this assert at 0 and update the `var countdown = new AsyncCountdownEvent(2);`  to a value of 2.

src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
```diff
@@ -132,6 +132,9 @@ public class WorkItemJob : IQueueJob<WorkItemData>, IHaveLogger {
 
                 return JobResult.Success;
             } catch (Exception ex) {
+
+                await ReportProgressAsync(handler, queueEntry, -1, $"Failed: {ex.Message}").AnyContext();
```
niemyjski, line 136:
```suggestion
            if (queueEntry.Value.SendProgressReports)
                await ReportProgressAsync(handler, queueEntry, -1, ex.Message).AnyContext();
```

We should probably respect the `SendProgressReports` property on the queue item. I just noticed there is another spot where we aren't looking at this `SendProgressReports` flag in the `progressCallback `.

@ejsmith any concerns around try catching this here? I don't really like nesting this but if this fails, we won't abandon or release locks.

ejsmith: We should respect the setting for sure. Try catch can't hurt.

ejsmith (changes requested):
Added some feedback

tests/Foundatio.Tests/Jobs/WorkItemJobTests.cs
akshay-zz: done

src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
akshay-zz: Added flag `SendProgressReports`

ejsmith (approved):
LGTM

niemyjski (approved):

niemyjski:
Thanks for the PR!

PR #290, opened 2023-09-04
Add write file stream

garcipat:
To write files in the file storage, usually you want to write to the stream directly, not pass the stream over as an argument. (as I SaveFileAsync()

This PR adds an overload where you can tell if you want to use the stream you get to read or to write only. 

There should be no breaking changes on existing things, but a new method is added, therefore implementations have to implement that.

### DIFF
--- a/src/Foundatio/Storage/FolderFileStorage.cs
+++ b/src/Foundatio/Storage/FolderFileStorage.cs
@@ -32,30 +32,46 @@ namespace Foundatio.Storage {
                 folder += Path.DirectorySeparatorChar;
 
             Folder = folder;
-            
+
             _logger.LogInformation("Creating {Directory} directory", folder);
             Directory.CreateDirectory(folder);
         }
 
-        public FolderFileStorage(Builder<FolderFileStorageOptionsBuilder, FolderFileStorageOptions> config) 
+        public FolderFileStorage(Builder<FolderFileStorageOptionsBuilder, FolderFileStorageOptions> config)
             : this(config(new FolderFileStorageOptionsBuilder()).Build()) { }
 
         public string Folder { get; set; }
         ISerializer IHaveSerializer.Serializer => _serializer;
 
-        public Task<Stream> GetFileStreamAsync(string path, CancellationToken cancellationToken = default) {
+        public Task<Stream> GetFileStreamAsync(string path, CancellationToken cancellationToken = default)
+            => GetFileStreamAsync(path, FileAccess.Read, cancellationToken);
+
+        public Task<Stream> GetFileStreamAsync(string path, FileAccess fileAccess, CancellationToken cancellationToken = default) {
             if (String.IsNullOrEmpty(path))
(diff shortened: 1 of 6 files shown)


src/Foundatio/Storage/FolderFileStorage.cs
```diff
@@ -32,31 +32,47 @@ public class FolderFileStorage : IFileStorage {
...
             : this(config(new FolderFileStorageOptionsBuilder()).Build()) { }
 
         public string Folder { get; set; }
         ISerializer IHaveSerializer.Serializer => _serializer;
 
-        public Task<Stream> GetFileStreamAsync(string path, CancellationToken cancellationToken = default) {
+        public Task<Stream> GetFileStreamAsync(string path, CancellationToken cancellationToken = default)
+            => GetFileStreamAsync(path, FileAccess.Read, cancellationToken);
+
+        public Task<Stream> GetFileStreamAsync(string path, FileAccess fileAccess, CancellationToken cancellationToken = default) {
```
ejsmith, line 49: I like this signature. I would say that we should mark the current method as deprecated so that we can get rid of it on the next major version.

ejsmith (changes requested):
Looks great! Thank you! Only bit of feedback is just to deprecate the old method so we can get rid of it since it is redundant now.

ejsmith:
Thank you @garcipat! Will you be able to create PRs for the other storage implementations as well?

garcipat:
What you mean? I went though all implementations of IFileStorage and added that. Or did I miss something?

ejsmith:
Yeah, not sure which ones you are using, but there are several `IFileStorage` implementations for S3, Azure Blob Storage, SshNet, Minio, Aliyun, and Redis. If you are using any of those and want to do PRs for those, that would be awesome. :-) If you are feeling really generous and want to go do PRs for all of them that would be crazy cool. :-)

garcipat:
Yes I rememeber. I means whats why we use fhe package. But I didnt realize its in other repositories. will see what I can do. Depends on how complicate it is with the other things. Also I dont know if I can test it with all of them. Will see. I will at least intend it.

garcipat:
> We probably want FileAccess fileAccess to default to Read, otherwise this is a pretty big breaking change. I upgraded a project and had quite a few compiler errors (warnings as errors on). The default for all of them was read. Thoughts?

Currently the implementation without the file access does use read to prevemt a breaking change. But we could set a default value if you would like to if we change it again anyway. 
I think it would be preferable to set it intentionally when changing from the obsolete method since its not obvious that the stream is only reading if nothing set. You only discover that when trying to write and it doesnt work.

 But thats personal preference I guess.

ejsmith:
It's not a breaking change because the old signature didn't have fileaccess on it and you would hit that one.

The new method requires FileAccess and it should continue to do so even when we remove the old one.

Also, @garcipat is going to update the enum to be a foundatio enum so that we control what stream modes are available. Which we said would just be read or write for now.

PR #292, opened 2023-09-23
Add storage streammode

garcipat:
Add package internal enumeration for stream mode that can be expanded as new options to use a stream is added.

The option to use FileAccess in FolderStorage was kept to be able to use it if desired and not cut off the functionality.


ejsmith (approved):
Looks good!

Issue #293, opened 2023-09-23
locks in async context


garcipat:
I discovered that in the file storage a lock object is used but the methods are actually asynchroneous.

Would it make sense to switch to a semaphore slim there instead?
see here https://github.com/FoundatioFx/Foundatio/blob/871be38c6b4b0ab19e86041f1f8714557428fbf8/src/Foundatio/Storage/FolderFileStorage.cs#L15

niemyjski:
Nice catch, we have an async lock we can use instead. Just search for AsyncLock. Would you mind submitting a pr for this?

garcipat:
I will check that then.

PR #294, opened 2023-09-23
Replace with async lock

garcipat:
In FolderFileStorage and InMemoryFileStorage, the lock wasn't async. Replaced it with the available AsyncLock recommended in this issue: #293 

There are two open points:
- Should the FolderFileStorage also use the async lock in all methods since its only used in Copy and Rename
- if the InMemoryFileStorage should use a [ConcurrentDictionary ](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=net-7.0) instead that is threadsafe by nature.

### DIFF
--- a/src/Foundatio/Storage/FolderFileStorage.cs
+++ b/src/Foundatio/Storage/FolderFileStorage.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Foundatio.AsyncEx;
 using Foundatio.Extensions;
 using Foundatio.Serializer;
 using Foundatio.Utility;
@@ -12,7 +13,7 @@ using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Foundatio.Storage {
     public class FolderFileStorage : IFileStorage {
-        private readonly object _lockObject = new();
+        private readonly AsyncLock _lock = new();
         private readonly ISerializer _serializer;
         protected readonly ILogger _logger;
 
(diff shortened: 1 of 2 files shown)


ejsmith (approved):
I like these changes. Nice work! Thanks @garcipat

garcipat:
@ejsmith what do you think about the two points I mentioned in the description?

ejsmith:
@garcipat sorry, missed those questions. I'm not sure about using locking in more places. Is it really needed? I can see using a lock on a rename operation, but I'm not convinced we need to be locking in other places. What are your thoughts?

garcipat:
> @garcipat sorry, missed those questions. I'm not sure about using locking in more places. Is it really needed? I can see using a lock on a rename operation, but I'm not convinced we need to be locking in other places. What are your thoughts?

@ejsmith I thought that it would be required as soon as you access files since you are maybe accessing the same file with two operations. I think the azure blob storage is doing that internally, but in the InMemory the lock is present in nearly every operation. So I thought its maybe necessary also in the FolderFileStorage. But was just a thing I saw while replacing the sync lock. Its probably not critical, otherwise it would have been reported earlier. 

What do you think about replacing the lock with a ConcurrentDictionary instead of having the lock everywhere?

ejsmith:
@garcipat I think it would be fine to do so. I'm not sure it's necessary, but I'd be good with that change if you want to do it.

garcipat:
I will see if I find time for it. It's definetly not urgent.

PR #302, opened 2024-07-19
Added QueueJob RunUntilEmptyAsync TimeSpan Overload, refactored cancellation tokens

niemyjski:
This should generally speed up tests where you don't want to wait the full 30 second default timeout.

### DIFF
--- a/src/Foundatio/Jobs/IQueueJob.cs
+++ b/src/Foundatio/Jobs/IQueueJob.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading;
+using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Foundatio.Queues;
 using Foundatio.Utility;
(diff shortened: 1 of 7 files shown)


src/Foundatio/Jobs/IQueueJob.cs
```diff
@@ -18,6 +19,19 @@ public interface IQueueJob<T> : IJob where T : class
...
+        CancellationToken cancellationToken = default) where T : class
+    {
+        if (acquireTimeout <= TimeSpan.Zero)
+            throw new ArgumentException("Acquire timeout must be greater than zero", nameof(acquireTimeout));
+
+        using var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        linkedCancellationTokenSource.CancelAfter(acquireTimeout);
+
+        // NOTE: This has to be awaited otherwise the linkedCancellationTokenSource cancel timer will not fire.
+        await job.RunUntilEmptyAsync(linkedCancellationTokenSource.Token);
```
niemyjski, line 32: NOTE: this was initially just returning the task, which did not dispose of the cancellation token but DID prevent the cancellation token timer from firing.... This was a huge time sink to track down.

src/Foundatio/Jobs/QueueJobBase.cs
```diff
@@ -34,17 +34,16 @@ public virtual async Task<JobResult> RunAsync(CancellationToken cancellationToke
     {
         IQueueEntry<T> queueEntry;
 
-        using (var timeoutCancellationTokenSource = new CancellationTokenSource(30000))
-        using (var linkedCancellationToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, timeoutCancellationTokenSource.Token))
+        using var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        linkedCancellationTokenSource.CancelAfter(TimeSpan.FromSeconds(30));
```
niemyjski, line 38: This seems more explicit about the behaviors and scoping of parent tokens.

src/Foundatio/Jobs/IQueueJob.cs
niemyjski: Might be something with https://github.com/microsoft/referencesource/blob/master/mscorlib/system/threading/CancellationTokenSource.cs#L460

ejsmith (approved):

PR #303, opened 2024-07-22
Approach for better implementation for RunUntilEmptyAsync

niemyjski:
@ejsmith I was curious if you would like this approach better as it's more than the total runtime has to be completed in X, and more first queued item if available has to finish in x time.

### DIFF
--- a/src/Foundatio/Jobs/IJob.cs
+++ b/src/Foundatio/Jobs/IJob.cs
@@ -30,14 +30,15 @@ public static class JobExtensions
         }
     }
 
-    public static async Task RunContinuousAsync(this IJob job, TimeSpan? interval = null, int iterationLimit = -1, CancellationToken cancellationToken = default, Func<Task<bool>> continuationCallback = null)
+    public static async Task<int> RunContinuousAsync(this IJob job, TimeSpan? interval = null, int iterationLimit = -1, CancellationToken cancellationToken = default, Func<Task<bool>> continuationCallback = null)
     {
         int iterations = 0;
         string jobName = job.GetType().Name;
         var logger = job.GetLogger();
+        bool isInformationLogLevelEnabled = logger.IsEnabled(LogLevel.Information);
 
         using var _ = logger.BeginScope(new Dictionary<string, object> { { "job", jobName } });
-        if (logger.IsEnabled(LogLevel.Information))
+        if (isInformationLogLevelEnabled)
             logger.LogInformation("Starting continuous job type {JobName} on machine {MachineName}...", jobName, Environment.MachineName);
 
         while (!cancellationToken.IsCancellationRequested)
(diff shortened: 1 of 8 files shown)


src/Foundatio/Jobs/IQueueJob.cs
```diff
@@ -19,6 +19,10 @@ public interface IQueueJob<T> : IJob where T : class
 
 public static class QueueJobExtensions
 {
+    /// <summary>
+    /// Will run wait for the acquire timeout to expire waiting if there are no queued items. It will then run until the queue is empty.
+    /// NOTE: The acquire timeout will not be reset until after the first job is processed,
+    /// </summary>
     public static async Task RunUntilEmptyAsync<T>(this IQueueJob<T> job, TimeSpan acquireTimeout,
```
ejsmith, line 26: Maybe rename from acquireTimeout to waitTimeout.

src/Foundatio/Jobs/IQueueJob.cs
```diff
@@ -27,9 +31,29 @@ public static class QueueJobExtensions
 
         using var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
         linkedCancellationTokenSource.CancelAfter(acquireTimeout);
+        bool hasAcquireTimeout = true;
+
+        var logger = job.GetLogger();
 
         // NOTE: This has to be awaited otherwise the linkedCancellationTokenSource cancel timer will not fire.
-        await job.RunUntilEmptyAsync(linkedCancellationTokenSource.Token);
+        await job.RunContinuousAsync(cancellationToken: linkedCancellationTokenSource.Token, continuationCallback: async () =>
```
ejsmith, line 39: Why is this better? It seems like it's just repeating code.

niemyjski: It's slightly better because the current approach cancels RunUntilEmpty after x time so it's not really an acquire timeout but a max wait. This is more of an acquire as it will reset the cancel after the first item is processed, so it doesn't wait. The downside is the first item has to process within the acquire timeout.

ejsmith: I don't like that. It's called RunUntilEmpty. The only thing it should care about is when the queue is empty. And the timeout is just how long you are willing to wait for the queue to be empty.

niemyjski: So if it runs and gets cancelled should it throw if items are remaining or being processed?

ejsmith: Would be nice if it was able to return a count of items that were processed. I don't think it should throw.

niemyjski: Yeah, I know... I don't love it either. It really should really only be a max wait overall or max wait for acquiring first item... But the acquiring... not sure that is possible due to how it's written.

ejsmith: It should just do what it says. Wait X amount of time for the queue to be empty.

src/Foundatio/Jobs/IQueueJob.cs
```diff
@@ -19,24 +19,37 @@ public interface IQueueJob<T> : IJob where T : class
...
+    /// </summary>
+    /// <returns>The amount of queue items processed. This count will not be accurate if the job is cancelled.</returns>
+    public static async Task<int> RunUntilEmptyAsync<T>(this IQueueJob<T> job, CancellationToken cancellationToken = default) where T : class
     {
         var logger = job.GetLogger();
-        return job.RunContinuousAsync(cancellationToken: cancellationToken, continuationCallback: async () =>
+
+        // NOTE: processed count is not accurate if the continuation callback is skipped due to cancellation.
+        int processed = 0;
+        await job.RunContinuousAsync(cancellationToken: cancellationToken, continuationCallback: async () =>
```
niemyjski, line 49: this has an iteration count, but it will always be 1 for the queue job if cancelled immediately.

randylsu (approved):

src/Foundatio/Jobs/IJob.cs
```diff
@@ -72,18 +84,31 @@ public static async Task RunContinuousAsync(this IJob job, TimeSpan? interval =
...
 
-        logger.LogInformation("Finished continuous job type {JobName}: {IterationLimit} {Iterations}", jobName, Environment.MachineName, iterationLimit, iterations);
         if (cancellationToken.IsCancellationRequested && logger.IsEnabled(LogLevel.Trace))
             logger.LogTrace("Job cancellation requested");
 
-        if (logger.IsEnabled(LogLevel.Information))
-            logger.LogInformation("Stopping continuous job type {JobName} on machine {MachineName}...", jobName, Environment.MachineName);
+        if (isInformationLogLevelEnabled)
+        {
+            if (iterationLimit > 0)
```
niemyjski, line 98: Note this doesn't log out the processed count if queue job but I think that is fine.

src/Foundatio/Jobs/IQueueJob.cs
```diff
@@ -19,22 +19,31 @@ public interface IQueueJob<T> : IJob where T : class
 
 public static class QueueJobExtensions
 {
-    public static async Task RunUntilEmptyAsync<T>(this IQueueJob<T> job, TimeSpan acquireTimeout,
+    /// <summary>
+    /// Will run until the wait timeout expires, if expired the job will be cancelled.
```
ejsmith, line 23:
```suggestion
    /// Will run until the queue is empty or the wait time is exceeded.
```

ejsmith (approved):

PR #304, opened 2024-07-23
Various improvements to hosting scheduled jobs

ejsmith:
- Ability to add cron jobs using lambda expressions and not requiring a job class
- Ability to manage scheduled jobs at runtime


samples/Foundatio.HostingSample/Program.cs
```diff
@@ -90,7 +92,14 @@ public static IHostBuilder CreateHostBuilder(string[] args)
                     s.AddCronJob<EveryMinuteJob>("* * * * *");
 
                 if (evenMinutes)
-                    s.AddCronJob<EvenMinutesJob>("*/2 * * * *");
+                    s.AddCronJob("*/2 * * * *", async sp =>
+                    {
+                        var logger = sp.GetRequiredService<ILogger<Program>>();
+                        if (logger.IsEnabled(LogLevel.Information))
+                            logger.LogInformation("EvenMinuteJob Run Thread={ManagedThreadId}", Thread.CurrentThread.ManagedThreadId);
```
niemyjski, line 99: Wonder if we could inject a logger based on the job name string as a parameter. Then again you have the provider

src/Foundatio.Extensions.Hosting/Jobs/DynamicJob.cs
```diff
@@ -0,0 +1,26 @@
...
+    public DynamicJob(IServiceProvider serviceProvider, Func<IServiceProvider, CancellationToken, Task> action)
+    {
+        _serviceProvider = serviceProvider;
+        _action = action;
+    }
+
+    public async Task<JobResult> RunAsync(CancellationToken cancellationToken = default)
+    {
+        await _action(_serviceProvider, cancellationToken).AnyContext();
+
```
niemyjski, line 23: Should this be error handled with logging so we can return cancelled or failure with an error?

src/Foundatio.Extensions.Hosting/Jobs/JobHostExtensions.cs
```diff
@@ -55,18 +59,19 @@ public static IServiceCollection AddJob(this IServiceCollection services, Func<I
...
-            if (!services.Any(s => s.ServiceType == typeof(IHostedService) && s.ImplementationType == typeof(ScheduledJobService)))
-                services.AddTransient<IHostedService, ScheduledJobService>();
 
-            return services.AddTransient(s => new ScheduledJobRegistration(jobOptions.JobFactory ?? (s.GetRequiredService<T>), jobOptions.CronSchedule));
-        }
+        if (!services.Any(s => s.ServiceType == typeof(IHostedService) && s.ImplementationType == typeof(ScheduledJobService)))
+            services.AddTransient<IHostedService, ScheduledJobService>();
+
+        if (!services.Any(s => s.ServiceType == typeof(IScheduledJobManager) && s.ImplementationType == typeof(ScheduledJobManager)))
+            services.AddSingleton<ScheduledJobManager>();
```
niemyjski, line 72: Can we share any of this logic or move it to a private implementation / extension?

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobManager.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Threading.Tasks;
+using Foundatio.Caching;
+using Foundatio.Jobs;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Foundatio.Extensions.Hosting.Jobs;
+
+public interface IScheduledJobManager
+{
```
niemyjski, line 14: Seems like it would be nice to know if it's running or current state... one might take this into account before you update the schedule?

src/Foundatio.Extensions.Hosting/Jobs/JobHostExtensions.cs
```diff
@@ -33,18 +36,19 @@ public static IServiceCollection AddJob(this IServiceCollection services, Func<I
...
 
-            return services.AddTransient(s => new ScheduledJobRegistration(() => jobFactory(s), jobOptions.CronSchedule));
-        }
+        if (!services.Any(s => s.ServiceType == typeof(IHostedService) && s.ImplementationType == typeof(ScheduledJobService)))
+            services.AddTransient<IHostedService, ScheduledJobService>();
+
+        if (!services.Any(s => s.ServiceType == typeof(IScheduledJobManager) && s.ImplementationType == typeof(ScheduledJobManager)))
+            services.AddSingleton<ScheduledJobManager>();
+
+        return services.AddTransient(s => new ScheduledJobRegistration(jobOptions.CronSchedule, jobOptions.Name ?? Guid.NewGuid().ToString(), _ => jobFactory(s)));
```
niemyjski, line 51: Kind of feels like job name should be required. If you just use a guid, it really breaks the manager.

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobRunner.cs
```diff
@@ -0,0 +1,113 @@
...
+namespace Foundatio.Extensions.Hosting.Jobs;
+
+internal class ScheduledJobRunner
+{
+    private readonly Func<IServiceProvider, IJob> _jobFactory;
+    private readonly IServiceProvider _serviceProvider;
+    private CronExpression _cronSchedule;
+    private readonly ILockProvider _lockProvider;
+    private readonly ILogger _logger;
+    private readonly DateTime _baseDate = new(2010, 1, 1);
```
niemyjski, line 21: why 2010?

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobRunner.cs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private string _schedule;
+    public string Schedule
+    {
+        get { return _schedule;}
+        set
+        {
+            _cronSchedule = CronExpression.Parse(value);
+            NextRun = _cronSchedule.GetNextOccurrence(SystemClock.UtcNow);
```
niemyjski, line 58: Should you be able to change the schedule of anything in flight? What if they do a cron schedule every second?

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobRunner.cs
```diff
@@ -0,0 +1,113 @@
...
+        // using lock provider in a cluster with a distributed cache implementation keeps cron jobs from running duplicates
+        // TODO: provide ability to run cron jobs on a per host isolated schedule
+        return _lockProvider.TryUsingAsync(GetLockKey(NextRun.Value), t =>
+        {
+            // start running the job in a thread
+            RunTask = Task.Factory.StartNew(async () =>
+            {
+                var result = await _jobFactory(_serviceProvider).TryRunAsync(cancellationToken).AnyContext();
+                // TODO: Should we only set last run on success? Seems like that could be bad.
+                _logger.LogJobResult(result, JobName);
```
niemyjski, line 95: Seems like this would be an area for future improvement where you can see a history of successful runs / failures.

niemyjski: Perhaps a point where you could also log some metrics / diagnostics?

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobRunner.cs
```diff
@@ -0,0 +1,113 @@
...
+        }, TimeSpan.Zero, TimeSpan.Zero);
+    }
+
+    private string GetLockKey(DateTime date)
+    {
+        _cacheKeyPrefix ??= TypeHelper.GetTypeDisplayName(_jobFactory(_serviceProvider).GetType());
+
+        long minute = (long)date.Subtract(_baseDate).TotalMinutes;
+
+        return _cacheKeyPrefix + minute;
```
niemyjski, line 111: 👀

samples/Foundatio.HostingSample/Program.cs
ejsmith: No, not all jobs need a logger.

src/Foundatio.Extensions.Hosting/Jobs/JobHostExtensions.cs
ejsmith: It's like 4 lines of repeated code. Not worried about it.

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobManager.cs
ejsmith: Maybe, but I'll add it when someone actually needs it.

src/Foundatio.Extensions.Hosting/Jobs/JobHostExtensions.cs
ejsmith: It should already be always populated, but you don't need a name on every job and lots of jobs you would never update.

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobManager.cs
```diff
@@ -0,0 +1,124 @@
...
+    public void Remove(string jobName)
+    {
+        var job = Jobs.FirstOrDefault(j => j.JobName == jobName);
+        if (job != null)
+        {
+            Jobs.Remove(job);
+        }
+    }
+
+    internal List<ScheduledJobRunner> Jobs { get; } = new();
```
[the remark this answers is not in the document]
ejsmith: The class is singleton, but I suppose multiple people could be trying to add jobs at the same time.

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobRunner.cs
ejsmith: I didn't change this. It's the same as it was before.

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobRunner.cs
ejsmith: Cron schedules can only be every minute at the most.

src/Foundatio.Extensions.Hosting/Jobs/ScheduledJobRunner.cs
ejsmith: Adding metrics / spans would be good, but that has nothing to do with this change.

src/Foundatio.Extensions.Hosting/Cronos/CalendarHelper.cs
```diff
@@ -5,7 +5,7 @@
 using System;
 using System.Runtime.CompilerServices;
 
-namespace Cronos;
```
niemyjski, line 8: Not sure why we never did this, but would be nice to pull something like this in to help us upgrade this lib. https://github.com/exceptionless/Exceptionless.Net/blob/main/src/Exceptionless/update-json.ps1

ejsmith: It's only like 5 files and isn't likely to need to be updated.

PR #315, opened 2024-10-02
Added Async Disposable Action

niemyjski:
The only usage across our libs is in test base which kinda made me want to remove DisposibleAction.

The other usages are for repositories and hence this add. I don't love the name.

image

### DIFF
--- a/src/Foundatio/Utility/DisposableAction.cs
+++ b/src/Foundatio/Utility/DisposableAction.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading;
 
 namespace Foundatio.Utility;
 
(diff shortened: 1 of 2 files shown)


src/Foundatio/Utility/DisposableAction.cs
```diff
@@ -24,10 +24,7 @@ public DisposableAction(Action exitAction)
     /// </summary>
     void IDisposable.Dispose()
     {
-        if (_disposed)
-            return;
-
-        _exitAction();
-        _disposed = true;
+        var exitAction = Interlocked.Exchange(ref _exitAction, null);
```
ejsmith, line 27: Why add the locking here? I've never seen any dispose methods have locking in them. I'm pretty sure the .NET runtime isn't going to call dispose a bunch of times. I suppose you could call it manually a bunch of times in a row, but it still seems like overhead to add locking here.

PR #383, opened 2025-05-17
#382 tweak dependencies based on new tfm

### LINKED ISSUE #382: Add additional TFM to eliminate explicit dependencies

thompson-tomo:
Closes #382

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0;net8.0</TargetFrameworks>
     <Product>Foundatio</Product>
     <Description>Pluggable foundation blocks for building distributed apps.</Description>
     <PackageProjectUrl>https://github.com/FoundatioFx/Foundatio</PackageProjectUrl>
--- a/build/common.props
+++ b/build/common.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
     <Product>Foundatio</Product>
     <Description>Pluggable foundation blocks for building distributed apps.</Description>
     <PackageProjectUrl>https://github.com/FoundatioFx/Foundatio</PackageProjectUrl>
--- a/build/common.props
+++ b/build/common.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <Product>Foundatio</Product>
     <Description>Pluggable foundation blocks for building distributed apps.</Description>
     <PackageProjectUrl>https://github.com/FoundatioFx/Foundatio</PackageProjectUrl>
--- a/src/Foundatio.Xunit/Foundatio.Xunit.csproj
+++ b/src/Foundatio.Xunit/Foundatio.Xunit.csproj
@@ -7,8 +7,10 @@
     <ProjectReference Include="..\Foundatio\Foundatio.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.Collections.Immutable" Version="8.0" />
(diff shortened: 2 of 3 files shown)


ejsmith (changes requested):
https://github.com/dotnet/runtime/issues/115124#issuecomment-2836313581

This seems to say that we should just add the additional target frameworks and leave the packagereferences alone.

thompson-tomo:
So my reading of that issue is that recommendation only applies to bcl libraries & when the classes are exposed in public apis.

ejsmith:
My understanding is that if we just add the other targets that the dependencies will just automatically go away on platforms where those dependencies are already included.

thompson-tomo:
Not the case as they clearly mention the polyfil in particular the redirect. Just look at what the Azure team has done with their libraries where they kept the bcl but not STJ. See their summary in https://github.com/Azure/azure-sdk-for-net/issues/48977#issuecomment-2873290158

ejsmith:
Hmm. I think we are going to need to run the tests on each platform to make sure we didn't break anything.

build/common.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
niemyjski, line 4: Personally would rather skip 6 for now and just add the LTS, it's basically not supported too much longer.

niemyjski:
> Hmm. I think we are going to need to run the tests on each platform to make sure we didn't break anything.

Are we expecting to update all projects to target both of these? I'm guessing so or should we just run on lowest (netstandard)? Ideally someday I'd like to bump the netstandard for NRT.

build/common.props
thompson-tomo: I don't see the harm in having it there. Given there is no conditional compilation necessary and a small dependency advantage is gained by having it.

niemyjski: .NET 6, released in November 2021, reached its end of support on November 12, 2024

build/common.props
```diff
@@ -1,7 +1,7 @@
...
+    <TargetFrameworks>netstandard2.0;net6.0;net8.0</TargetFrameworks>
```
niemyjski, line 4:
```suggestion
    <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
```

build/common.props
ejsmith: Yeah, if it is out of support then I'd agree that we shouldn't add it.

src/Foundatio.Xunit/Foundatio.Xunit.csproj
```diff
@@ -7,8 +7,10 @@
     <ProjectReference Include="..\Foundatio\Foundatio.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.Collections.Immutable" Version="8.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0" />
     <PackageReference Include="xunit" Version="2.9.3" />
   </ItemGroup>
+  <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">
```
niemyjski, line 13: @thompson-tomo Do you think this should be bumped to net8 or track the lowest net target?

thompson-tomo: No it doesn't need to be as this is saying only for framework not net 6 or higher add this dependency.

src/Foundatio/Foundatio.csproj
```diff
@@ -1,10 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
+    <!-- Can be removed when lowest tfm is net 8 -->
     <PackageReference Include="Microsoft.Bcl.TimeProvider" Version="8.0" />
-    <PackageReference Include="System.Text.Json" Version="8.0.5" />
+    <!-- Can be removed when lowest tfm is net 6 -->
```
niemyjski, line 5:
```suggestion
    <!-- Can be removed when lowest TFM is net 6 -->
```

src/Foundatio/Foundatio.csproj
```diff
@@ -1,10 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <ItemGroup>
+    <!-- Can be removed when lowest tfm is net 8 -->
```
niemyjski, line 3:
```suggestion
    <!-- Can be removed when lowest TFM is net 8 -->
```

niemyjski:
Thanks for the PR. I'm going to fix up some of the new warnings and build pipeline so all TFMs run for tests

PR #395, opened 2025-07-24
BREAKING: Improved Cache Tests and normalized ICacheClient.RemoveByPrefix

niemyjski:
Remove By Prefix

1. If it's null or empty we remove all cache items.
2. we escape the entire prefix and check cache keys for values, this means whitespace and other special characters are respected.

This increases coverage especially for hybrid cache client and remove by prefix.

Also introduces CacheException (there are probably more places we should be throwing it, especially in the InMemoryCacheClient)


src/Foundatio.TestHarness/Caching/CacheClientTestsBase.cs
```diff
@@ -429,24 +429,319 @@ public virtual async Task CanRemoveByPrefixAsync()
...
+            await cache.SetAsync($"similar{pattern}test", 3);
+            await cache.SetAsync($"not{pattern.Replace("*", "X")}test", 4);
+
+            Assert.Equal(2, await cache.RemoveByPrefixAsync(pattern));
+            Assert.False(await cache.ExistsAsync($"{pattern}test1"));
+            Assert.False(await cache.ExistsAsync($"{pattern}test2"));
+            Assert.True(await cache.ExistsAsync($"similar{pattern}test"));
+            Assert.True(await cache.ExistsAsync($"not{pattern.Replace("*", "X")}test"));
+        }
+    }
```
niemyjski, line 532: @ejsmith wildcards here needs some discussion, main reason tests are failing.

src/Foundatio.TestHarness/Caching/HybridCacheClientTestBase.cs
```diff
@@ -0,0 +1,190 @@
...
+    public HybridCacheClientTestBase(ITestOutputHelper output) : base(output)
+    {
+        _distributedCache = new InMemoryCacheClient(o => o.LoggerFactory(Log));
+        _messageBus = new InMemoryMessageBus(o => o.LoggerFactory(Log));
+    }
+
+    /// <summary>
+    /// Returns a hybrid cache client that has a distributed shared cache. For in memory tests, this will be overridden with a shared cache.
+    /// </summary>
+    protected virtual HybridCacheClient GetDistributedHybridCacheClient(bool shouldThrowOnSerializationError = true)
```
niemyjski, line 28: I don't love this, but this is needed for the in memory hybrid tests that must have a shared distributed client... Also not sure if it makes sense that this class has an instanced of in memory services in the ctor.

src/Foundatio.TestHarness/Caching/HybridCacheClientTestBase.cs
```diff
@@ -0,0 +1,190 @@
...
+        Assert.Equal("value", (await secondCache.GetAsync<string>(cacheKey)).Value);
+        Assert.Equal(1, secondCache.LocalCache.Count);
+
+        Assert.Equal(1, await firstCache.RemoveByPrefixAsync("test-"));
+        Assert.Equal(0, firstCache.InvalidateCacheCalls);
+        Assert.Equal(0, firstCache.LocalCache.Count);
+
+        await Task.Delay(250); // Allow time for local cache to clear
+        Assert.Equal(1, secondCache.InvalidateCacheCalls);
+        Assert.Equal(0, secondCache.LocalCache.Count);
```
niemyjski, line 182: ensure the message bus is invalidating.

src/Foundatio/Caching/InMemoryCacheClient.cs
```diff
@@ -183,18 +183,23 @@ public Task<int> RemoveAllAsync(IEnumerable<string> keys = null)
 
     public Task<int> RemoveByPrefixAsync(string prefix)
     {
-        var keysToRemove = new List<string>();
         string normalizedPrefix = String.IsNullOrWhiteSpace(prefix) ? "*" : prefix.Trim();
-        var regex = new Regex(String.Concat("^", Regex.Escape(normalizedPrefix.Contains("*") ? normalizedPrefix : $"{normalizedPrefix}*"), "$").Replace("\\*", ".*?"));
+        if (String.Equals(normalizedPrefix, "*"))
+            return RemoveAllAsync();
```
niemyjski: short circuit a bunch of allocations and checking below.

src/Foundatio/Caching/InMemoryCacheClient.cs
```diff
@@ -183,18 +183,23 @@ public Task<int> RemoveAllAsync(IEnumerable<string> keys = null)
...
+
+        var keysToRemove = new List<string>();
+
         try
         {
+            var regex = new Regex(String.Concat("^", Regex.Escape(normalizedPrefix.EndsWith("*") ? normalizedPrefix : $"{normalizedPrefix}*"), "$").Replace("\\*", ".*?"));
```
niemyjski: moved this into the try so if it throws it's logged. Also changed the regex to endsWith so it better handles prefixes.

src/Foundatio/Caching/ScopedCacheClient.cs
```diff
@@ -70,11 +70,17 @@ protected string GetScopedCacheKey(string unscopedKey)
 
     public Task<bool> RemoveAsync(string key)
     {
+        if (String.IsNullOrEmpty(key))
```
niemyjski, line 73: Added argument validation here as tests started failing when I added scoped tests. There was no checks here and so when we passed the unscoped client arguments it was always with a value so it didn't work the same way.

src/Foundatio/Caching/CacheException.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Foundatio.Caching;
+
+public class CacheException : Exception
```
niemyjski, line 5: @ejsmith not sure what to call this. We have StorageException. But what should the others be MessagingException or MessageBusException, CachingException, LockException etc..

niemyjski (commented):
Self reviewed

src/Foundatio/Caching/CacheException.cs
niemyjski: @ejsmith any preference on this before we cut a release and we ship this

PR #402, opened 2025-08-13
Adds queue dequeue activity tracking

niemyjski:
Introduces activity tracking for queue dequeue operations to monitor performance and diagnose issues.

Adds tags to the activity with relevant information such as queue name, job id, entry type, id, correlation id, and attempts.

Improves the dequeue experience around viewing span data as you don't know some spans take a long time, in this case sqs queues doing long polling.

image

### DIFF
--- a/src/Foundatio/Jobs/QueueJobBase.cs
+++ b/src/Foundatio/Jobs/QueueJobBase.cs
@@ -141,7 +141,7 @@ public abstract class QueueJobBase<T> : IQueueJob<T>, IHaveLogger, IHaveLoggerFa
 
     protected virtual Activity StartDequeueActivity()
     {
-        var activity = FoundatioDiagnostics.ActivitySource.StartActivity("DequeueQueueEntry", ActivityKind.Internal);
+        var activity = FoundatioDiagnostics.ActivitySource.StartActivity("DequeueQueueEntry");
         if (activity is null)
             return null;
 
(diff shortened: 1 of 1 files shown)


src/Foundatio/Jobs/QueueJobBase.cs
```diff
@@ -137,13 +139,40 @@ public async Task<JobResult> ProcessAsync(IQueueEntry<T> queueEntry, Cancellatio
...
+        if (activity is null || !activity.IsAllDataRequested)
+            return;
+
+        if (entry is null)
+            return;
+
+        activity.AddTag("EntryType", entry.EntryType.FullName);
+        activity.AddTag("Id", entry.Id);
+        activity.AddTag("CorrelationId", entry.CorrelationId);
+        activity.AddTag("Attempts", entry.Attempts.ToString());
```
niemyjski: @ejsmith not sure if we should add attempts here

src/Foundatio/Jobs/QueueJobBase.cs
```diff
@@ -161,8 +190,9 @@ protected virtual void EnrichProcessQueueEntryActivity(Activity activity, IQueue
         activity.AddTag("EntryType", entry.EntryType.FullName);
         activity.AddTag("Id", entry.Id);
         activity.AddTag("CorrelationId", entry.CorrelationId);
+        activity.AddTag("Attempts", entry.Attempts.ToString());
```
niemyjski: @ejsmith not sure if we should add attempts here

PR #426, opened 2025-11-25
[Cache]: Added GetAllExpirationsAsync and SetAllExpirationsAsync. Also refactored tests and improved test coverage. 

niemyjski:
There are one or two small breaking changes around argument validation where we are now enforcing you cannot pass in a null dictionary, and we are now asserting keys cannot be null or empty (white space is valid). This enforces consistency across our caching implementations (redis) as well as all cache method implementations that already took dictionaries and had null checks.

### DIFF
--- a/src/Foundatio.TestHarness/Caching/CacheClientTestsBase.cs
+++ b/src/Foundatio.TestHarness/Caching/CacheClientTestsBase.cs
@@ -14,7 +14,7 @@ using Xunit.Abstractions;
 
 namespace Foundatio.Tests.Caching;
 
-public abstract partial class CacheClientTestsBase : TestWithLoggingBase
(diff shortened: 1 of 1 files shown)


src/Foundatio.TestHarness/Caching/CacheClientTestsBase.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Foundatio.Tests.Caching;
 
-public abstract class CacheClientTestsBase : TestWithLoggingBase
+public abstract partial class CacheClientTestsBase : TestWithLoggingBase
```
niemyjski:
```suggestion
public abstract class CacheClientTestsBase : TestWithLoggingBase
```

niemyjski:
@randylsu @twehner One step closer

Issue #429, opened 2025-12-16
Telemetry Actions not marked as Errors


dviry:
Maybe I am missing something - but shouldn't Telemetry Actions be set to Status error when... an error occurs?

In my code (in this case a QueueJob handler based on QueueJobBase<>) I had to wrap as following:

```c#
protected async override Task<JobResult> ProcessQueueEntryAsync(QueueEntryContext<IdpWorkItem> context)
{
   try
   {
      ......
   }
  catch (Exception ex)
  {
      Activity.Current?.SetStatus(ActivityStatusCode.Error, ex.Message);
      Activity.Current?.RecordException(ex);
      return JobResult.FromException(ex);
   }
}
```

Wouldn't be better to have that in here: https://github.com/FoundatioFx/Foundatio/blob/ae95423657e8d83f1b584992685291db92c986f5/src/Foundatio/Jobs/QueueJobBase.cs#L137

Or even better here:
https://github.com/FoundatioFx/Foundatio/blob/ae95423657e8d83f1b584992685291db92c986f5/src/Foundatio/Jobs/JobRunner.cs#L213

I can find multiple places where this could be done (IMO the higher the better) - but is there any reason this was not done originally?

Thanks

niemyjski:
This is the first time I've come across these api's personally. How is this working out for you, are you able to see each queue attempt failure as an error? What happens when it succeeds after a retry? I think as long as you are marking a non root span as an error you are fine.

What other places would you set this?

niemyjski:
I guess from my perspective it would be great to surface the errors, I'm just wondering what we would get for free by logging errors and you'd think that would surface them. Also if it is a retry and completes it technically had error spans but finished partial success. Do you find this works well? We would love a pr if you could otherwise I could take a stab at this but sounds like you have more experience with this :-).

As far as messaging goes, I think we still would love to do #274, we have limited resources and lots of projects / apps we work on, most could benefit from it. We have recently introduced some newer foundatio libraries and it would be nice to see some interop on the handler side and play a bit better with kafka. With our friend Opus 4.5 I think it's way more likey projects like that get across the finish line quicker. We are always welcoming feedback and contributions. Feel free to join our discord!

niemyjski:
@dviry I've created a pr to address this will try to push this out with 13-beta2. Would love any feedback. https://github.com/FoundatioFx/Foundatio/pull/446

PR #446, opened 2026-02-06
Add OpenTelemetry error recording with Activity.AddException polyfill

niemyjski:
## Summary

Addresses #429 - telemetry activities were not consistently marked as errors when exceptions occurred.

- **Polyfill \Activity.AddException\** for net8 that matches the .NET 10 API signature and semantics (internal class, won't leak). On .NET 10+ the built-in method takes over automatically via conditional compilation.
- **\SetErrorStatus\ extension** that both sets \ActivityStatusCode.Error\ with a descriptive message AND records the exception as an \ActivityEvent\ with \xception.type\, \xception.message\, and \xception.stacktrace\ tags per OpenTelemetry semantic conventions.
- Uses \xception.ToString()\ for the stacktrace tag to capture the full exception chain including inner exceptions.
- **Consistent error recording** across all activity scopes:
  - Job runner (\RunContinuousAsync\) records errors from \JobResult\ including the exception when present
  - \QueueJobBase\ records errors on dequeue, process, and catch activities
  - \WorkItemJob\ records errors on deserialization failures, missing handlers, handler failures, and unhandled exceptions
  - \MessageBusBase\ records errors on subscriber handler failures
  - \JobWithLockBase\ records errors when lock acquisition throws
  - \QueueBase\ records errors when queue stats collection fails
  - \CacheLockProvider\ records errors when lock acquisition fails (non-cancellation)
- Lock cancellation log level changed from Trace to Debug for better visibility
- Tests follow 3-part naming convention with AAA pattern, using a single \ThrowingJob\ that validates inner exception chain capture

## Test plan

- [x] All 1783 existing tests pass
- [x] New telemetry tests validate error status, exception event tags, and inner exception stacktrace capture
- [x] Builds clean on both net8.0 and net10.0 targets with 0 warnings

### DIFF
--- a/src/Foundatio/Jobs/QueueJobBase.cs
+++ b/src/Foundatio/Jobs/QueueJobBase.cs
@@ -121,8 +121,6 @@ public abstract class QueueJobBase<T> : IQueueJob<T>, IHaveLogger, IHaveLoggerFa
             }
             else
             {
-                activity?.SetErrorStatus(result.Error, result.Message ?? result.Error?.Message);
-
                 if (result.Error != null || result.Message != null)
                     _logger.LogError(result.Error, "{QueueName} queue entry {QueueEntryId} returned an unsuccessful response: {Message}", _queueName, queueEntry.Id, result.Message ?? result.Error?.Message);
 
--- a/src/Foundatio/Queues/SharedQueueOptions.cs
+++ b/src/Foundatio/Queues/SharedQueueOptions.cs
@@ -65,7 +65,8 @@ public class SharedQueueOptionsBuilder<T, TOptions, TBuilder> : SharedOptionsBui
     {
         ArgumentNullException.ThrowIfNull(behavior);
 
-        Target.Behaviors ??= new List<IQueueBehavior<T>>();
+        if (Target.Behaviors == null)
+            Target.Behaviors = new List<IQueueBehavior<T>>();
         Target.Behaviors.Add(behavior);
 
         return (TBuilder)this;


src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
```diff
@@ -76,7 +76,9 @@ public async Task<JobResult> ProcessAsync(IQueueEntry<WorkItemData> queueEntry,
         if (workItemDataType == null)
         {
             await queueEntry.AbandonAsync().AnyContext();
-            return JobResult.FailedWithMessage($"Abandoning {queueEntry.Value.Type} work item: {queueEntry.Id}: Could not resolve work item data type");
+            var result = JobResult.FailedWithMessage($"Abandoning {queueEntry.Value.Type} work item: {queueEntry.Id}: Could not resolve work item data type");
+            Activity.Current?.SetErrorStatus(message: result.Message);
```
niemyjski: Feels like the job runner should handle this.

src/Foundatio/Jobs/WorkItemJob/WorkItemJob.cs
```diff
@@ -145,6 +148,8 @@ await Task.WhenAll(
 
             if (!workItemContext.Result.IsSuccess)
             {
+                activity?.SetErrorStatus(workItemContext.Result.Error, workItemContext.Result.Message ?? workItemContext.Result.Error?.Message);
+
```
ejsmith: Not sure we want this. This is a normal state where the work item returned that it wasn't a success and that can happen for any number of reasons. I don't think we want our telemetry spans to be errored in this situation.

src/Foundatio/Jobs/IJob.cs
```diff
@@ -88,6 +89,9 @@ public static async Task<int> RunContinuousAsync(this IJob job, JobOptions optio
             var result = await job.TryRunAsync(cancellationToken).AnyContext();
             logger.LogJobResult(result, options.Name);
 
+            if (!result.IsSuccess && !result.IsCancelled)
+                activity?.SetErrorStatus(result.Error, result.Message ?? result.Error?.Message);
+
```
ejsmith: Same here

src/Foundatio/Lock/CacheLockProvider.cs
```diff
@@ -183,9 +183,14 @@ public async Task<ILock> AcquireAsync(string resource, TimeSpan? timeUntilExpire
             _lockTimeoutCounter.Add(1);
 
             if (cancellationToken.IsCancellationRequested)
-                _logger.LogTrace("Cancellation requested for lock {Resource} ({LockId}) after {Duration:g}", resource, lockId, sw.Elapsed);
+            {
+                _logger.LogDebug("Cancellation requested for lock {Resource} ({LockId}) after {Duration:g}", resource, lockId, sw.Elapsed);
+            }
             else
+            {
+                activity?.SetErrorStatus(message: $"Failed to acquire lock {resource} after {sw.Elapsed:g}");
```
ejsmith: Not sure about this one either. We need to be careful to not fill our telemetry up with a bunch of errors that really aren't errors. Not being able to acquire a lock is a normal thing and the lock returns and says it didn't get the lock.

src/Foundatio/Jobs/IJob.cs
niemyjski: Feels like we would want to log here. the job returned a failed state.

ejsmith: Don't say anything about changing logging. Just not setting activity error for something that possibly isn't an error.

niemyjski: meant setting the error status on the activity. you had a failed job state the span should be an error no?

src/Foundatio/Jobs/IJob.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics;
```
niemyjski:
```suggestion
```

PR #508, opened 2026-04-29
Make AcquireAsync throw on failure to fix the silent-null lock footgun

ejsmith:
## Why

`ILockProvider.AcquireAsync` previously returned `Task<ILock?>` — `null` when the lock couldn't be obtained. Callers who forgot the null check silently ran their protected work **without holding the lock**, the exact scenario the lock was meant to prevent. We hit this in a downstream caller (an account-number generation lock) where a forgotten null check would have allowed duplicate numbers under contention.

The fix is to make the dangerous shape impossible to write by accident: failing to acquire should not produce a value of type `ILock`.

## What

`ILockProvider` now exposes two acquisition shapes as interface methods, sharing a private core in each implementation so neither path pays an exception-as-control-flow cost:

| API | Returns | Use when |
| --- | --- | --- |
| `AcquireAsync` | `Task<ILock>` — throws `LockAcquisitionTimeoutException` on failure | The work cannot run safely without the lock (the safer default). |
| `TryAcquireAsync` | `Task<ILock?>` — `null` on failure | Lock unavailability is normal control flow (best-effort dedupe, opportunistic work). |

`CacheLockProvider` and `ThrottlingLockProvider` route both public methods through a private `TryAcquireCoreAsync`. `ScopedLockProvider` forwards each shape to the corresponding method on the unscoped provider.

`LockAcquisitionTimeoutException` derives from a new `LockException` base (mirroring `CacheException` / `StorageException`) so consumers can `catch (LockException)` to handle all lock-related failures uniformly.

## Source / behavior compatibility

This is a breaking change for callers — both at the source level and at the behavior level. Three categories to be aware of:

**1. Code that *calls* `AcquireAsync` and assigns to `var` / `ILock` / `ILock?`** — recompiles fine. The only signature shift is the return type going from `Task<ILock?>` to `Task<ILock>`. Existing null checks become dead code (compiler warns "always non-null"); switch them to `try`/`catch (LockAcquisitionTimeoutException)` if you want to handle the failure, or move to `TryAcquireAsync` if null-on-failure is what you want.

**2. Code that *returns* `AcquireAsync`'s result from a method typed `Task<ILock?>`** — gets a compile error. `Task<T>` is a class and is **invariant** in `T`, so `Task<ILock>` is not assignable to `Task<ILock?>`. Concrete known case: `Foundatio.Repositories`'s `ReindexWorkItemHandler.GetWorkItemLockAsync`. Fix is either to retype the method as `Task<ILock>` (and let the throw propagate) or switch the call site to `TryAcquireAsync`.

**3. Code that handled the null return as graceful skip** — silently changes from "skip the work" to "throw `LockAcquisitionTimeoutException`". Known examples in `Foundatio.Repositories`: `ElasticConfiguration` and `MigrationManager` both have `if (lock is null) return …` paths. These need to switch to `TryAcquireAsync` to preserve the existing skip behavior.

The internal best-effort callers in this repo (`ScheduledJobInstance`, `WithLockingJob`, `ThrottledJob`, `SampleQueueJob`, `QueueTestBase`) have already been migrated to `TryAcquireAsync` as part of this PR.

External Foundatio repos that ship a *provider* (`Foundatio.Redis`, `Foundatio.AWS`, …) consume `CacheLockProvider` rather than implementing `ILockProvider` directly, so the new interface methods don't require changes there.

## Cancelled-token contract

Both `AcquireAsync` and `TryAcquireAsync` treat an already-cancelled `cancellationToken` as an immediate-deadline attempt: a single acquire is tried, and on failure `AcquireAsync` throws `LockAcquisitionTimeoutException` while `TryAcquireAsync` returns `null`. This preserves the existing `acquireTimeout: TimeSpan.Zero` try-once-no-wait pattern (used by `TryUsingAsync(..., TimeSpan.Zero)` in `LockTestBase.DoLockedWorkAsync`).

## Other changes

- New `LockException` base type (in `Foundatio.Lock`), and `LockAcquisitionTimeoutException` derived from it.
- The default exception message is neutral (`Failed to acquire lock for resource 'X'.`) so it covers cancellation and multi-resource cases as well as plain timeouts.
- Multi-resource throwing `AcquireAsync` overloads materialize `resources` once so single-use enumerables aren't walked twice and the exception message reflects the actual attempt.
- Convenience overloads' XML docs explicitly note that a null `acquireTimeout` applies a 30-second default (pre-existing behavior, now documented).
- Updated `docs/guide/locks.md` to document the dual API and recommend `AcquireAsync` as the safer default.
- New `LockTestBase` tests covering the throwing shape (single resource, cancelled token, multi-resource).

## Test plan

- [x] `dotnet build Foundatio.slnx` — clean
- [x] `dotnet test Foundatio.slnx` — 1861 passed, 13 skipped (pre-existing skips), 0 failed
- [ ] Verify CI passes
- [ ] Follow-up: migrate downstream `Foundatio.Repositories` callers (`ReindexWorkItemHandler`, `ElasticConfiguration`, `MigrationManager`, `CustomFieldDefinitionRepository`)

### DIFF
--- a/docs/guide/locks.md
+++ b/docs/guide/locks.md
@@ -9,9 +9,9 @@ Locks ensure a resource is only accessed by one consumer at any given time. Foun
 ```csharp
 public interface ILockProvider
 {
-    Task<ILock?> AcquireAsync(string resource, TimeSpan? timeUntilExpires = null,
-                               bool releaseOnDispose = true,
-                               CancellationToken cancellationToken = default);
+    Task<ILock?> TryAcquireAsync(string resource, TimeSpan? timeUntilExpires = null,
+                                  bool releaseOnDispose = true,
+                                  CancellationToken cancellationToken = default);
     Task<bool> IsLockedAsync(string resource);
     Task ReleaseAsync(string resource, string lockId);
     Task ReleaseAsync(string resource);
@@ -30,6 +30,17 @@ public interface ILock : IAsyncDisposable
 }
 ```
 
+### Choosing Between `AcquireAsync` and `TryAcquireAsync`
+
+`ILockProvider` exposes two acquisition shapes:
+
+| API | Returns | Use when |
+| --- | --- | --- |
+| `TryAcquireAsync` | `Task<ILock?>` — `null` on failure | Lock unavailability is a normal control-flow outcome (best-effort dedupe, opportunistic work) |
+| `AcquireAsync` | `Task<ILock>` — throws `LockAcquisitionTimeoutException` on failure | The work cannot safely run without the lock — failure is genuinely exceptional |
+
+The throwing `AcquireAsync` is provided as an extension method on `ILockProvider`. Pick whichever one matches the caller's intent — they share the same underlying acquisition logic.
+
 ## Implementations
 
 ### CacheLockProvider
@@ -47,8 +58,8 @@ var cache = new InMemoryCacheClient();
 var messageBus = new InMemoryMessageBus();
 var locker = new CacheLockProvider(cache, messageBus);
 
-await using var lck = await locker.AcquireAsync("my-resource");
-if (lck != null)
+await using var lck = await locker.TryAcquireAsync("my-resource");
+if (lck is not null)
 {
     // Exclusive access to resource
     await DoExclusiveWorkAsync();
@@ -80,8 +91,8 @@ var throttledLocker = new ThrottlingLockProvider(
 );
 
 // Only allows 10 operations per minute across all instances
-var lck = await throttledLocker.AcquireAsync("api-rate-limit");
-if (lck != null)
+var lck = await throttledLocker.TryAcquireAsync("api-rate-limit");
+if (lck is not null)
 {
     await CallExternalApiAsync();
     await lck.ReleaseAsync();
@@ -106,7 +117,7 @@ var baseLock = new CacheLockProvider(cache, messageBus);
 var tenantLock = new ScopedLockProvider(baseLock, "tenant:abc");
 
 // Lock key becomes: "tenant:abc:resource-1"
-await using var lck = await tenantLock.AcquireAsync("resource-1");
+await using var lck = await tenantLock.TryAcquireAsync("resource-1");
 ```
 
 ## Basic Usage
@@ -116,10 +127,10 @@ await using var lck = await tenantLock.AcquireAsync("resource-1");
 ```csharp
 var locker = new CacheLockProvider(cache, messageBus);
 
-// Acquire lock
-var lck = await locker.AcquireAsync("my-resource");
+// Best-effort acquire — null when held elsewhere
+var lck = await locker.TryAcquireAsync("my-resource");
 
-if (lck != null)
+if (lck is not null)
 {
     try
     {
@@ -139,8 +150,8 @@ if (lck != null)
 The recommended pattern uses `await using` for automatic release:
 
 ```csharp
-await using var lck = await locker.AcquireAsync("my-resource");
-if (lck != null)
+await using var lck = await locker.TryAcquireAsync("my-resource");
+if (lck is not null)
 {
     // Lock is automatically released when scope ends
     await DoExclusiveWorkAsync();
@@ -149,11 +160,11 @@ if (lck != null)
 
 ### Non-Blocking Acquire
 
-Check if lock was acquired:
+Try to acquire without waiting:
 
 ```csharp
-await using var lck = await locker.AcquireAsync("my-resource");
-if (lck == null)
+await using var lck = await locker.TryAcquireAsync("my-resource");
+if (lck is null)
 {
     // Resource is locked by another process
     return;
@@ -165,28 +176,38 @@ await DoWorkAsync();
 
 ### Blocking Acquire with Timeout
 
-Wait for lock with cancellation:
+Wait up to a timeout, throwing on failure:
 
 ```csharp
-using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
-
 try
 {
     await using var lck = await locker.AcquireAsync(
         "my-resource",
-        cancellationToken: cts.Token
-    );
+        acquireTimeout: TimeSpan.FromSeconds(30));
 
-    if (lck != null)
-    {
-        await DoWorkAsync();
-    }
+    await DoWorkAsync();
 }
-catch (OperationCanceledException)
+catch (LockAcquisitionTimeoutException)
 {
-    // Timeout waiting for lock
+    // Lock could not be acquired before the timeout elapsed.
     _logger.LogWarning("Timed out waiting for lock");
 }
+```
+
+If the failure is expected control flow, prefer `TryAcquireAsync` so you don't pay for an exception:
+
+```csharp
+await using var lck = await locker.TryAcquireAsync(
+    "my-resource",
+    acquireTimeout: TimeSpan.FromSeconds(30));
+
+if (lck is null)
+{
+    _logger.LogWarning("Could not acquire lock");
+    return;
+}
+
+await DoWorkAsync();
 ```
 
 ## Lock Expiration
@@ -197,7 +218,7 @@ Locks expire automatically to prevent deadlocks:
 
 ```csharp
 // Lock expires after 5 minutes
-await using var lck = await locker.AcquireAsync(
+await using var lck = await locker.TryAcquireAsync(
     "my-resource",
     timeUntilExpires: TimeSpan.FromMinutes(5)
 );
@@ -208,12 +229,12 @@ await using var lck = await locker.AcquireAsync(
 For long-running operations, renew the lock:
 
 ```csharp
-await using var lck = await locker.AcquireAsync(
+await using var lck = await locker.TryAcquireAsync(
     "my-resource",
     timeUntilExpires: TimeSpan.FromMinutes(1)
 );
 
-if (lck != null)
+if (lck is not null)
 {
     // Do some work
     await DoPartOneAsync();
@@ -231,8 +252,8 @@ if (lck != null)
 For very long operations, set up automatic renewal:
 
 ```csharp
-await using var lck = await locker.AcquireAsync("my-resource");
-if (lck == null) return;
+await using var lck = await locker.TryAcquireAsync("my-resource");
+if (lck is null) return;
 
 using var cts = new CancellationTokenSource();
 
@@ -265,9 +286,9 @@ Ensure only one instance processes a resource:
 ```csharp
 public async Task ProcessOrderAsync(int orderId)
 {
-    await using var lck = await _locker.AcquireAsync($"order:{orderId}");
+    await using var lck = await _locker.TryAcquireAsync($"order:{orderId}");
 
-    if (lck == null)
+    if (lck is null)
     {
         _logger.LogInformation("Order {OrderId} is being processed elsewhere", orderId);
         return;
@@ -287,9 +308,9 @@ public async Task RunAsLeaderAsync(CancellationToken ct)
 {
     while (!ct.IsCancellationRequested)
     {
-        await using var lck = await _locker.AcquireAsync("leader:job-runner");
+        await using var lck = await _locker.TryAcquireAsync("leader:job-runner");
 
-        if (lck != null)
+        if (lck is not null)
         {
             _logger.LogInformation("This instance is now the leader");
 
@@ -315,16 +336,12 @@ public async Task RunAsLeaderAsync(CancellationToken ct)
 ```csharp
 public async Task<Order> CreateOrderAsync(CreateOrderRequest request)
 {
-    // Prevent duplicate orders for same customer
+    // Prevent duplicate orders for same customer; throw if we can't lock.
     await using var lck = await _locker.AcquireAsync(
         $"create-order:{request.CustomerId}",
-        timeUntilExpires: TimeSpan.FromSeconds(30)
+        timeUntilExpires: TimeSpan.FromSeconds(30),
+        acquireTimeout: TimeSpan.FromSeconds(5)
     );
-
-    if (lck == null)
-    {
-        throw new ConcurrencyException("Another order is being created");
-    }
 
     // Check for recent duplicates
     var recentOrder = await _db.GetRecentOrderAsync(request.CustomerId);
@@ -380,9 +397,9 @@ public class RateLimitedApiClient
 
     public async Task<T> GetAsync<T>(string endpoint)
     {
-        await using var lck = await _throttler.AcquireAsync("external-api");
+        await using var lck = await _throttler.TryAcquireAsync("external-api");
 
-        if (lck == null)
+        if (lck is null)
         {
             throw new RateLimitExceededException();
         }
@@ -399,9 +416,9 @@ public class RateLimitedApiClient
 public async Task<IActionResult> ProcessRequest(string userId)
 {
     // 10 requests per minute per user
-    await using var lck = await _throttler.AcquireAsync($"user:{userId}:api");
+    await using var lck = await _throttler.TryAcquireAsync($"user:{userId}:api");
 
-    if (lck == null)
+    if (lck is null)
     {
         return StatusCode(429, "Too many requests");
     }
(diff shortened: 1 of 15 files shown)


docs/guide/locks.md
```diff
@@ -9,9 +9,9 @@ Locks ensure a resource is only accessed by one consumer at any given time. Foun
 ```csharp
 public interface ILockProvider
 {
-    Task<ILock?> AcquireAsync(string resource, TimeSpan? timeUntilExpires = null,
-                               bool releaseOnDispose = true,
-                               CancellationToken cancellationToken = default);
+    Task<ILock?> TryAcquireAsync(string resource, TimeSpan? timeUntilExpires = null,
```
niemyjski, line 15: Try assumes an out parameter, but I guess you are saying it's null when cannot be acquired but it's almost like it would return boolean, this is also a major breaking change.

ejsmith: Good call on both concerns — going with this design instead: `AcquireAsync` stays on the interface (no rename, source-compatible at the call site — `Task<ILock?>` → `Task<ILock>` is a covariant tightening, existing callers just recompile). It now throws `LockAcquisitionTimeoutException` on failure, so the footgun goes away because there's no null branch to forget. `TryAcquireAsync` is added as a sibling interface method for callers who *want* null on failure. Both share a private core in each implementation, so we don't pay an exception-as-control-flow cost on the best-effort path.

src/Foundatio/Lock/ILockProvider.cs
```diff
@@ -117,20 +122,73 @@ public static Task RenewAsync(this ILockProvider provider, ILock @lock, TimeSpan
...
+    }
+
+    /// <summary>
+    /// Tries to acquire the lock, waiting up to <paramref name="acquireTimeout"/> before giving up.
+    /// Returns <c>null</c> if the timeout elapses without acquiring the lock.
+    /// </summary>
+    public static async Task<ILock?> TryAcquireAsync(this ILockProvider provider, string resource, TimeSpan? timeUntilExpires = null, TimeSpan? acquireTimeout = null)
+    {
+        using var cancellationTokenSource = acquireTimeout.ToCancellationTokenSource(TimeSpan.FromSeconds(30));
+        return await provider.TryAcquireAsync(resource, timeUntilExpires, true, cancellationTokenSource.Token).AnyContext();
```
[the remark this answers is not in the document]
ejsmith: Documented the implicit 30s default explicitly in the XML doc on the `acquireTimeout` overloads (`A null acquireTimeout applies a 30-second default.`). Behavior is unchanged from the original code; the docs no longer disagree with the code.

src/Foundatio/Lock/ILockProvider.cs
```diff
@@ -117,20 +122,73 @@ public static Task RenewAsync(this ILockProvider provider, ILock @lock, TimeSpan
...
+    /// <summary>
+    /// Acquires the lock, waiting up to <paramref name="acquireTimeout"/> before throwing
+    /// <see cref="LockAcquisitionTimeoutException"/>.
+    /// </summary>
+    /// <exception cref="LockAcquisitionTimeoutException">The lock could not be acquired before the timeout elapsed.</exception>
+    public static async Task<ILock> AcquireAsync(this ILockProvider provider, string resource, TimeSpan? timeUntilExpires = null, TimeSpan? acquireTimeout = null)
     {
         using var cancellationTokenSource = acquireTimeout.ToCancellationTokenSource(TimeSpan.FromSeconds(30));
-        return await provider.AcquireAsync(resource, timeUntilExpires, true, cancellationTokenSource.Token).AnyContext();
+        var l = await provider.TryAcquireAsync(resource, timeUntilExpires, true, cancellationTokenSource.Token).AnyContext();
```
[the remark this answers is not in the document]
ejsmith: Documented the implicit 30s default explicitly in the XML doc on the `acquireTimeout` overloads (`A null acquireTimeout applies a 30-second default.`). Behavior is unchanged from the original code; the docs no longer disagree with the code.

src/Foundatio/Lock/ILockProvider.cs
```diff
@@ -288,21 +350,61 @@ public static Task<bool> TryUsingAsync(this ILockProvider locker, string resourc
...
+    // AcquireAsync — multiple resources. Throws on failure.
+    // ------------------------------------------------------------------
+
+    /// <exception cref="LockAcquisitionTimeoutException">One or more locks could not be acquired.</exception>
+    public static async Task<ILock> AcquireAsync(this ILockProvider provider, IEnumerable<string> resources, TimeSpan? timeUntilExpires = null, bool releaseOnDispose = true, CancellationToken cancellationToken = default)
+    {
+        var l = await provider.TryAcquireAsync(resources, timeUntilExpires, releaseOnDispose, cancellationToken).AnyContext();
+        if (l is null)
+            throw new LockAcquisitionTimeoutException(String.Join(",", resources));
+
```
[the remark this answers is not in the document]
ejsmith: Fixed. The throwing multi-resource `AcquireAsync` overloads now materialize `resources` once (`resources as string[] ?? resources.ToArray()`) and use the materialized list for both the acquisition call and the exception message. `ArgumentNullException.ThrowIfNull(resources)` added at the top.

src/Foundatio/Lock/ThrottlingLockProvider.cs
```diff
@@ -43,7 +43,7 @@ public ThrottlingLockProvider(ICacheClient cacheClient, int maxHitsPerPeriod = 1
     TimeProvider IHaveTimeProvider.TimeProvider => _timeProvider;
     IResiliencePolicyProvider IHaveResiliencePolicyProvider.ResiliencePolicyProvider => _resiliencePolicyProvider;
 
-    public async Task<ILock?> AcquireAsync(string resource, TimeSpan? timeUntilExpires = null, bool releaseOnDispose = true, CancellationToken cancellationToken = default)
+    public async Task<ILock?> TryAcquireAsync(string resource, TimeSpan? timeUntilExpires = null, bool releaseOnDispose = true, CancellationToken cancellationToken = default)
     {
         _logger.LogTrace("AcquireLockAsync: {Resource}", resource);
```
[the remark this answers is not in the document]
ejsmith: Fixed — the trace now says `"TryAcquireAsync: {Resource}"`. The acquisition logic in `ThrottlingLockProvider` lives in a private `TryAcquireCoreAsync` shared between the public `AcquireAsync` and `TryAcquireAsync` methods.

src/Foundatio/Lock/LockAcquisitionTimeoutException.cs
```diff
@@ -0,0 +1,39 @@
...
+/// Callers that treat lock unavailability as a normal control-flow outcome (best-effort
+/// dedupe, opportunistic work) should call <c>TryAcquireAsync</c> instead and check the
+/// returned <see cref="ILock"/> for <c>null</c>.
+/// </remarks>
+public sealed class LockAcquisitionTimeoutException : Exception
+{
+    public LockAcquisitionTimeoutException(string resource)
+        : base($"Failed to acquire lock for resource '{resource}' before the timeout elapsed.")
+    {
+        Resource = resource;
```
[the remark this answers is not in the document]
ejsmith: Fixed. Default message is now `Failed to acquire lock for resource 'X'.` — neutral about cause, so it's correct for plain `CancellationToken` cancellation as well as timeouts, and for multi-resource as well as single. Updated the `Resource` property's xmldoc to call out that it can be a comma-separated list for multi-resource acquisition.

docs/guide/locks.md
niemyjski: Should this derive from LockException? so it aligns with CacheException, StorageException?

src/Foundatio/Lock/ILockProvider.cs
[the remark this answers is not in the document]
niemyjski: What would you do if you passed in a cancelled lock and couldn't acquire, guessing this would throw immediately.

niemyjski (commented):
## Review

The design is sound and well-motivated — the null-return footgun is a real production risk, and splitting into `AcquireAsync` (throws) / `TryAcquireAsync` (nullable) is the right fix. The implementation is clean: sharing `TryAcquireCoreAsync` avoids exception-as-control-flow costs, all internal callers are migrated, docs are thorough, and review feedback has been addressed well. A few concerns below.

---

### 1. `Task<T>` Is Invariant — Downstream Repos Will Get Compile Errors, Not Just Behavioral Changes

The PR description claims source compatibility because `Task<ILock?>` → `Task<ILock>` is a "covariant tightening." **This is incorrect** — `Task<T>` is a class and is **invariant** in `T`. Code that explicitly returns `Task<ILock?>` from a method typed as returning `Task<ILock?>` will get a **compile error** when calling `AcquireAsync` (which now returns `Task<ILock>`).

Concrete example in **Foundatio.Repositories**:

```csharp
// ReindexWorkItemHandler.cs — return type is Task<ILock?>
public override Task<ILock?> GetWorkItemLockAsync(...)
{
    return _lockProvider.AcquireAsync(...); // Task<ILock> is NOT assignable to Task<ILock?>
}
```

This will be a **compile error**, not a silent behavioral change.

Additionally, callers that handle the null return gracefully will now throw instead:

- `ElasticConfiguration.cs` — calls `AcquireAsync(...)` then `if (configLock is null)` to skip gracefully → post-PR this throws instead of skipping
- `MigrationManager.cs` — calls `AcquireAsync(..., TimeSpan.Zero)` then `if (migrationsLock == null) return MigrationResult.UnableToAcquireLock` → dead code, now throws
- `CustomFieldDefinitionRepository.cs` — was already throwing on null, so this one is fine (just changes exception type)

**Recommendation:** A coordinated migration tracking issue for downstream repos would be helpful. The PR description should also correct the "covariant tightening" claim — `Task<T>` invariance means this is a harder break than described.

---

### 2. Exception Hierarchy: Should `LockAcquisitionTimeoutException` Derive from a Base `LockException`?

Echoing @niemyjski's comment — other Foundatio domains have domain-specific base exceptions (`CacheException`, `StorageException`, `MessageBusException`). Deriving from bare `Exception` is inconsistent. A `LockException` base (even if introduced in this PR) would let consumers write `catch (LockException)` to handle all lock-related failures uniformly.

---

### 3. Section Divider Comments

The PR adds block comments like:

```csharp
// ------------------------------------------------------------------
// AcquireAsync — single resource. Throws LockAcquisitionTimeoutException
// ------------------------------------------------------------------
```

Per AGENTS.md: "No code comments unless necessary — code should be self-explanatory." The XML doc summaries on each method already convey the same information. These section headers are arguably noise that could drift out of sync. Consider removing them and relying on the XML docs alone.

src/Foundatio/Lock/ILockProvider.cs
```diff
@@ -117,20 +145,67 @@ public static Task RenewAsync(this ILockProvider provider, ILock @lock, TimeSpan
         return provider.RenewAsync(@lock.Resource, @lock.LockId, timeUntilExpires);
     }
 
-    public static Task<ILock?> AcquireAsync(this ILockProvider provider, string resource, TimeSpan? timeUntilExpires = null, CancellationToken cancellationToken = default)
+    // ------------------------------------------------------------------
```
niemyjski: These section divider comments should be removed — the XML doc summaries on each method already convey the same information. Having both means two places to keep in sync, and the block comments will inevitably drift. Let the XML docs do their job.

randylsu:
This silent failure burned as well, so I'm happy to see this change. Looks good to me.

docs/guide/locks.md
ejsmith: Done in 55436db8 — added a `LockException` base mirroring `CacheException` / `StorageException`, and `LockAcquisitionTimeoutException` now derives from it. Consumers can `catch (LockException)` to handle all lock failures uniformly.

src/Foundatio/Lock/ILockProvider.cs
ejsmith: Done in 55436db8 — all six section-divider blocks removed. The XML doc summaries on each method already say what each group is for.

src/Foundatio/Lock/ILockProvider.cs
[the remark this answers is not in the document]
ejsmith: Confirmed in 01c3042c, with a clarifying note added to the XML docs. An already-cancelled token is treated as an immediate-deadline attempt — same semantics callers already get via `acquireTimeout: TimeSpan.Zero`. The provider does one acquire attempt; on failure `AcquireAsync` throws `LockAcquisitionTimeoutException` and `TryAcquireAsync` returns null.

We considered making pre-cancelled tokens throw `OperationCanceledException` at entry, but that would break the existing `TryUsingAsync(..., acquireTimeout: TimeSpan.Zero)` try-once-no-wait pattern (used in `LockTestBase.DoLockedWorkAsync`). Treating cancelled-at-entry as "immediate deadline" composes naturally with that idiom and lines up with what you guessed.

src/Foundatio/Lock/ThrottlingLockProvider.cs
```diff
@@ -43,9 +43,23 @@ public ThrottlingLockProvider(ICacheClient cacheClient, int maxHitsPerPeriod = 1
...
+    public async Task<ILock> AcquireAsync(string resource, TimeSpan? timeUntilExpires = null, bool releaseOnDispose = true, CancellationToken cancellationToken = default)
+    {
+        var l = await TryAcquireCoreAsync(resource, timeUntilExpires, releaseOnDispose, cancellationToken).AnyContext();
+        if (l is null)
+            throw new LockAcquisitionTimeoutException(resource);
+
+        return l;
+    }
+
+    private async Task<ILock?> TryAcquireCoreAsync(string resource, TimeSpan? timeUntilExpires, bool releaseOnDispose, CancellationToken cancellationToken)
```
niemyjski:
```suggestion
    private async Task<ILock?> TryAcquireImplAsync(string resource, TimeSpan? timeUntilExpires, bool releaseOnDispose, CancellationToken cancellationToken)
```

I think we use Impl for most things, especially on queues. But this works as is but more consistent.

src/Foundatio/Lock/CacheLockProvider.cs
```diff
@@ -97,7 +97,21 @@ private Task OnLockReleasedAsync(CacheLockReleased msg, CancellationToken cancel
...
+    public async Task<ILock> AcquireAsync(string resource, TimeSpan? timeUntilExpires = null, bool releaseOnDispose = true, CancellationToken cancellationToken = default)
+    {
+        var l = await TryAcquireCoreAsync(resource, timeUntilExpires, releaseOnDispose, cancellationToken).AnyContext();
+        if (l is null)
+            throw new LockAcquisitionTimeoutException(resource);
+
+        return l;
+    }
+
+    private async Task<ILock?> TryAcquireCoreAsync(string resource, TimeSpan? timeUntilExpires, bool releaseOnDispose, CancellationToken cancellationToken)
```
niemyjski:
```suggestion
    private async Task<ILock?> TryAcquireImplAsync(string resource, TimeSpan? timeUntilExpires, bool releaseOnDispose, CancellationToken cancellationToken)
```

src/Foundatio/Lock/ILockProvider.cs
```diff
@@ -288,21 +351,67 @@ public static Task<bool> TryUsingAsync(this ILockProvider locker, string resourc
...
     {
         using var cancellationTokenSource = acquireTimeout.ToCancellationTokenSource(TimeSpan.FromSeconds(30));
-        return await provider.AcquireAsync(resources, timeUntilExpires, releaseOnDispose, cancellationTokenSource.Token).AnyContext();
+        return await provider.TryAcquireAsync(resources, timeUntilExpires, releaseOnDispose, cancellationTokenSource.Token).AnyContext();
+    }
+
+    /// <exception cref="LockAcquisitionTimeoutException">One or more locks could not be acquired.</exception>
+    public static async Task<ILock> AcquireAsync(this ILockProvider provider, IEnumerable<string> resources, TimeSpan? timeUntilExpires = null, bool releaseOnDispose = true, CancellationToken cancellationToken = default)
+    {
+        ArgumentNullException.ThrowIfNull(resources);
```
niemyjski, line 369: what if the resources count is 0

src/Foundatio/Lock/ThrottlingLockProvider.cs
ejsmith: Renamed in 4c8ab6b2.

src/Foundatio/Lock/CacheLockProvider.cs
ejsmith: Renamed in 4c8ab6b2.

src/Foundatio/Lock/ILockProvider.cs
ejsmith: Already a no-op success — the multi-resource `TryAcquireAsync` extension returns `EmptyLock.Empty` when `resourceList.Length is 0` (line 282-283), and the throwing `AcquireAsync` overloads here just delegate to it, so they get back a non-null `EmptyLock` and skip the throw. No change needed.

niemyjski (approved):

